Issue #691, opened 2026-08-18
ann_rf added to 11 metrics targets and to none of the registry unit maps — every registry writer aborts, hidden by error = "continue"


JohnGavin:
`ann_rf` was added to 11 metrics targets and to none of the 11 registry unit maps — every registry writer aborts, and `error = "continue"` hid it

## What happened

#677 slice 4 (#686) added an `ann_rf` column to every strategy's `*_metrics`
target so QA gate S17 could assert `sharpe == (cagr - ann_rf) / vol`. It did not
add `ann_rf` to any of the hand-maintained `units` maps the registry writers pass
to `historicaldata::hd_metric_record()`.

`hd_metric_record()` correctly refuses to write a metric with no declared unit —
that guard is #640 working exactly as designed:

```
✖ Missing metric_unit for metric "ann_rf".
ℹ Every metric written to `bt.metric` must declare a unit.
ℹ Allowed units: "fraction", "percent", "ratio", "count", "days", and "years".
```

So every registry writer whose strategy gained `ann_rf` now aborts.

## Why it was invisible

Each writer selects its metric columns **by exclusion**:

```r
metric_cols <- setdiff(names(full_row), "period")     # inclusion set: everything
ltr_units   <- c(months = "count", cagr = "percent", ...)   # inclusion list: hand-typed
```

Adding one column to a metrics target therefore *automatically* enters the
value set and *does not* enter the unit set. The two lists are maintained by
opposite mechanisms, so they drift by construction, not by accident.

`error = "continue"` in `docs/_targets.R` then made `tar_make()` exit 0 with
these targets errored, and `tar_read()` kept serving each writer's last good
value. Nothing in `scripts/verify.sh` runs a target body, so full verification
passed on every one of the four commits since #686.

It surfaced only because the post-merge check added by #680
(`scripts/check_pipeline_errors.R`, PR #690) was run against a real build —
which is the first thing that check has caught.

## Blast radius

Confirmed errored in the build at `6b71b6b`:

- `ltr_register_runs`
- `avoid_worst_register_runs`
- `fm_register_runs`
- `tom_register_runs`
- `cmr_registry_run`
- `drif_register_runs`

Statically identical and expected to error on their next rebuild —
each has `ann_rf` in its metrics target and no `ann_rf` in its unit map:

- `xgb_drif_register_runs`, `stk_max_register_runs`, `stk_drif_register_runs`,
[...]

Issue #677, opened 2026-08-17
leaderboard 'sharpe' mixes three different formulas — LTR ranks top on a statistic only it and Risk State use


JohnGavin:
The `sharpe` column on the `leaderboard` target is computed by **three different formulas** depending on which plan file produced the row. Strategies are ranked against each other on it, and #635's sizing reads it.

Found while verifying the #665 borrow change (PR #676): a sweep computing `cagr/vol` disagreed with the published Sharpe by an amount that varied per strategy in a way no single risk-free rate could explain.

## The three formulas

All three verified by reading source, not inferred:

| # | formula | strategies | source |
|---|---|---|---|
| **A** | `(ann_ret - mean(rf_ret) * 12) / ann_vol` — geometric, risk-free adjusted | Factor MAX, Factor DRIF | `R/plan_factormax.R:187`, `R/plan_drif.R:293` |
| **B** | `cagr / vol` — geometric, **no risk-free deduction** | Managed Futures, Value (HML) | `R/plan_managed_futures.R:199,271`, `R/plan_ev_ebit.R:121,192` |
| **C** | `hac_sharpe` — **naive arithmetic** HAC Sharpe, a different statistic | LTR, Risk State | `R/plan_leaderboard.R:74` (`.norm_ltr`), `R/plan_leaderboard.R:137` (`.norm_rsc`), both `rename(sharpe = hac_sharpe)` |

## The evidence

Implied risk-free rate per strategy — the rate that would reconcile each published `sharpe` with its own published `cagr` and `vol`, i.e. `(cagr − sharpe × vol)`. Computed from the `leaderboard` target at `dd0992a`:

| strategy | cagr | vol | sharpe | implied rf | family |
|---|---|---|---|---|---|
| **LTR** | 7.70 | 17.20 | **2.360** | **−32.89%** | C |
| Factor MAX | 3.49 | 8.92 | −0.098 | +4.36% | A |
| Factor DRIF | 5.15 | 10.31 | 0.076 | +4.36% | A |
| CMR | −0.79 | 5.04 | −0.524 | +1.85% | A-like |
| Stock MAX | −14.00 | 17.51 | −0.899 | +1.74% | A-like |
| Stock DRIF | −18.21 | 14.73 | −1.332 | +1.41% | A-like |
| XGB DRIF | −20.10 | 17.05 | −1.261 | +1.41% | A-like |
| PSO Optimal | 0.85 | 6.57 | −0.084 | +1.41% | A-like |
[...]

Issue #665, opened 2026-08-06
financing is modelled on neither side consistently: 4 of 17 strategies carry borrow cost, none earns securities-lending income


JohnGavin:
Broader than #664, which only asks to make the ambiguity *visible*. This asks for the modelling itself: **every short leg must carry a borrow cost, and every long leg should carry the securities-lending income it would earn.**

### Borrow cost — modelled for 4 of 17 strategies

| strategy | short leg | borrow modelled |
|---|---|---|
| Stock MAX, Stock DRIF, XGB DRIF, LTR | yes | yes — flat `0.03` annual |
| **Mom Pre-Peak, Mom Post-Peak, Mom 12-2** | **yes, 100% short leg** | **no** |
| CMR | yes (`weight = -1/n_short`) | no |
| Factor MAX, Factor DRIF, TOM, Risk State, Avoid Worst, OLMAR-1 | no | N/A, correctly |

`R/plan_mom_prepeak.R:338` computes `ret_ls = ret_long - ret_short - 2*cost_per_trade` — trading cost only, no financing. Their published CAGR and Sharpe are **overstated by an unquantified amount**.

### Securities-lending income — modelled nowhere, for anything

A grep across `R/` and `packages/historicaldata/R/` for `lend`, `rebate`, `lending revenue` returns **zero hits**. No strategy earns anything on the long book it holds, even though a long equity portfolio is a lendable asset and lending revenue is a real, routinely-material component of a market-neutral book's P&L.

## Why modelling one side only is its own problem

The two omissions bias in **opposite** directions:

| what is omitted | effect on reported return |
|---|---|
| borrow cost on shorts | **overstates** |
| lending income on longs | **understates** |

So the current state is not simply "conservative" or "optimistic" — it is **inconsistent across strategies**:

- The 3 mom_prepeak variants omit both → partial self-cancellation, net direction unknown
- The 4 with borrow modelled omit only lending income → understated relative to a full model
- Long-only strategies omit lending income → understated

Strategies are therefore not comparable on the leaderboard for a reason unrelated to skill, which is exactly the class of defect #637/#640/#663 keep surfacing.

## The flat 0.03 is itself a problem

Where borrow *is* modelled it is a hardcoded `0.03` annual (`R/plan_stock_backtest.R:392-393`, `R/plan_ltr_momentum.R:27-28`) with no source and no provenance beyond a `cost_source_ref` pointing at the line that hardcodes it.

Two objections:

[...]

PR #692, opened 2026-08-18
fix(registry): add ann_rf to 11 unit maps; fix mom_prepeak percent bug (#691)

JohnGavin:
## Summary

`ann_rf` was added to 11 metrics targets by #677 slice 4 (#686) but never
added to the hand-maintained `<prefix>_units` maps the registry writers pass
to `historicaldata::hd_metric_record()`. That function correctly aborts on
any column with no declared unit (#640), so every writer whose metrics
target gained `ann_rf` began erroring — hidden by `error = "continue"` in
`docs/_targets.R`.

## Part 1 — unit assigned per map, and the evidence

Each unit was independently verified against the metrics-computing
function's own source, not copied from the issue's table. All 11 checked
out against the issue except `pso_units` (see the discrepancy noted below).

| Map (file) | Unit added | Evidence |
|---|---|---|
| `ltr_units` (`R/plan_ltr_momentum.R`) | `ann_rf = "percent"` | `compute_ltr_metrics()` returns `round(sr$ann_rf * 100, 1)` |
| `aw_units` (`R/plan_avoid_worst.R`) | `ann_rf = "percent"` | `metrics_for()` returns `round(sr$ann_rf * 100, 2)` |
| `tom_units` (`R/plan_turn_of_month.R`) | `ann_rf_tom = "percent"` | `calc()` returns `round(sr_s$ann_rf * 100, 2)`; column is named `ann_rf_tom`, not `ann_rf` |
| `rsc_units` (`R/plan_risk_state.R`) | `ann_rf = "percent"` | `calc_metrics()` returns `round(sr_result$ann_rf * 100, 2)` |
| `mom_prepeak_units` (`R/plan_mom_prepeak.R`) | `ann_rf = "percent"` | `m$ann_rf <- round(.mom_prepeak_ann_rf(...) * 100, 2)` (×3 call sites) |
| `cmr_units` (`R/plan_commodities_mean_reversion.R`) | `ann_rf = "fraction"` | `.compute_cmr_metrics()` returns `round(sr$ann_rf, 4)` — never ×100 |
| `fm_units` (`R/plan_factormax.R`) | `ann_rf = "fraction"` | `calc_metrics()`: `ann_rf <- mean(df$rf_ret) * 12` — never ×100 |
| `drif_units` (`R/plan_drif.R`) | `ann_rf = "fraction"` | `calc_metrics()`: `ann_rf <- mean(df$rf_ret, na.rm=TRUE) * 12` — never ×100 |
| `stk_units` (`R/plan_stock_backtest.R`) | `ann_rf = "fraction"` | shared `calc_backtest_metrics()`: `rf_ann <- mean(df[[rf_col]]) * 12` — never ×100 |
| `xgb_units` (`R/plan_xgb_signal.R`) | `ann_rf = "fraction"` | same shared `calc_backtest_metrics()` as `stk` |
| `pso_units` (`R/plan_portfolio_opt.R`) | `ann_rf = "fraction"` | `calc_port_metrics()`: `ann_rf = rf_ann` (`mean(df$rf_ret) * 12`) — never ×100 |

**Discrepancy found in Part 1:** the issue lists `pso_optimal_register_runs`
among the six writers "expected to error on their next rebuild." That's not
correct. `.pso_optimal_register_runs()` selects its metric columns via an
explicit `intersect(c("months","opt_cagr","opt_vol","opt_sharpe",
"opt_maxdd"), names(full_row))` allow-list — unlike the other ten writers,
which all use a `setdiff(names(full_row), <excluded>)` exclusion pattern that
automatically pulls in any new column. `port_metrics` does carry `ann_rf`,
but `pso_cols` never selects it, so this specific target does not currently
attempt to write `ann_rf` and was not erroring. I added `ann_rf = "fraction"`
to `pso_units` anyway (matches the issue's literal instruction and is
forward-compatible/consistent with the other ten maps), but left `pso_cols`
unchanged — expanding what this writer publishes to the registry is a
separate, larger decision than fixing a unit-declaration bug. Documented
inline at `R/plan_portfolio_opt.R:537-549`.

Confirmed `mom_prepeak_gauntlet_register`/`gauntlet_units` is unaffected, per
the issue — it writes a disjoint metric set (`hac_sharpe`, `ff5_alpha`,
`pbo`, …) that never gained `ann_rf`.

## Part 2 — mom_prepeak percent/fraction correction

`mom_prepeak_units` declared `cagr`/`vol`/`max_dd` as `"fraction"`. Verified
wrong: `.mom_prepeak_compute_metrics()`
(`packages/historicaldata/R/utils_mom_prepeak_metrics.R:128-130`) returns
`round(cagr * 100, 1)` / `round(vol * 100, 1)` / `round(max_dd * 100, 1)` —
percent. `R/plan_leaderboard.R`'s `.norm_mom_sibling()` independently
confirms this by dividing `cagr`/`vol`/`max_dd`/`ann_rf` by 100 before
adding the row to the leaderboard. Changed all three (plus the new `ann_rf`)
to `"percent"`.

## Part 3 — a test that fails against pre-fix code

Added `tests/testthat/test-registry-unit-map-coverage.R`. Since the 11 unit
maps are locals inside each writer's helper function (not importable
directly), the test statically parses each plan file's AST (`parse()`, no
target-body evaluation, no store build — a worktree must not build the
store) to extract the literal `<prefix>_units <- c(...)` assignment, and
asserts its names cover the metrics target's verified numeric column set
(hardcoded per-map, cited against the exact source lines read for this PR).
A separate test asserts `mom_prepeak_units$cagr/vol/max_dd == "percent"`.

**Demonstrated genuine failure against pre-fix code:** temporarily restored
`R/plan_ltr_momentum.R` to its pre-fix (`HEAD`) content and re-ran the test
file:

```
FAILURE: 'test-registry-unit-map-coverage.R:174:9' ----------
Expected `missing` to equal `character(0)`.
Differences:
`actual`:   "ann_rf"
`expected`:

ltr_units (plan_ltr_momentum.R) is missing unit declarations for: ann_rf

[ FAIL 1 | WARN 0 | SKIP 0 | PASS 14 ]
```

Restored the fix and re-ran — all 15 pass:

```
[ FAIL 0 | WARN 0 | SKIP 0 | PASS 15 ]
```

Per `snapshot-test-policy`, the new `cli_abort()` inside the test's
AST-extraction helper has `expect_snapshot(error = TRUE, ...)` coverage
(`tests/testthat/_snaps/registry-unit-map-coverage.md`). The snapshot
message uses `basename(file)`, not the full path, so it stays portable
across checkouts/CI (`portable-build-artifacts`).

## Part 4 — scripts/verify.sh

Run in the foreground (backgrounded automatically past the 10-minute Bash
timeout on the cold-shell entry; waited for completion, then re-ran once
more after a snapshot-portability fix to get final fresh evidence):

```
PASS: _targets.R parses
PASS: docs/_targets.R parses
PASS: docs/_targets.R is structurally valid (tar_validate, 7.65s)

SKIP count this run: 0

SKIP count this run: 15

```

Root suite: `total=587 failed=3 skipped=0` (matches baseline exactly — the
15 new tests all pass and are counted in the 587 total). Package suite:
`total=695 failed=1 skipped=15` (matches baseline exactly). No regressions.

`tar_make()` was **not** run — a worktree must not build the targets store
(would race the main checkout's store, per `.claude/CLAUDE.md`). The six
previously-errored registry targets (`ltr_register_runs`,
`avoid_worst_register_runs`, `fm_register_runs`, `tom_register_runs`,
`cmr_registry_run`, `drif_register_runs`) plus the five statically-identical
ones (`xgb_drif_register_runs`, `stk_max_register_runs`,
`stk_drif_register_runs`, `mom_prepeak_register_runs`, `rsc_register_runs`)
need a real rebuild to confirm green — that's for the maintainer to run
against the main checkout.

## Follow-up (out of scope for this PR)

- Backfill or invalidate the `bt.metric` rows already written under
  `mom_prepeak`'s previously-wrong `"fraction"` unit for `cagr`/`vol`/
  `max_dd` — those rows are stored as raw percent values (e.g. `-21.0`) but
  were labelled `"fraction"`, so any reader that trusted the label got a
  value 100× too large. This PR does not touch existing `bt.metric` rows.
- The issue's "structural half" (lifting all 12 maps to named top-level
  constants + a `tar_make()`-time QA gate) is explicitly out of scope here —
  this PR's test (Part 3) is a static-analysis substitute that catches the
  same class of drift without a store build, per the issue's own fallback
  option (b).

## Not closing the issue

This PR does not use a closing keyword. The issue's "structural half" fix
(lifting the maps to constants + a QA gate, so the *next* new column can't
reproduce this silently) is explicitly left as follow-up, and the
`bt.metric` backfill for the mom_prepeak percent-bug rows is also
unaddressed. Both are named directly in the issue as required or
recommended; leaving the issue open until a maintainer decides how (or
whether) to pursue them, per #665.

### DIFF
--- a/R/plan_avoid_worst.R
+++ b/R/plan_avoid_worst.R
@@ -1270,6 +1270,8 @@ plan_avoid_worst <- function() {
   # Units (#640): aw_metrics stores cagr/vol/max_dd as PERCENT (x*100 in
   # metrics_for() above — the source #637 flagged as percent-native), sharpe
   # is a scale-free ratio, n_days is a count, and years is a year-duration.
+  # ann_rf (#677 slice 4, #691) is PERCENT, same convention as cagr
+  # (round(sr$ann_rf * 100, 2) above).
   full_row <- aw_metrics[
     aw_metrics$period == "Full Period" & aw_metrics$scenario == "All Days",
     , drop = FALSE
@@ -1278,7 +1280,8 @@ plan_avoid_worst <- function() {
     metric_cols <- setdiff(names(full_row), c("period", "scenario"))
     aw_units <- c(
       years = "years", n_days = "count", cagr = "percent",
-      vol = "percent", max_dd = "percent", sharpe = "ratio"
+      vol = "percent", max_dd = "percent", sharpe = "ratio",
+      ann_rf = "percent"
     )
     historicaldata::hd_metric_record(
       con, uu, full_row[, metric_cols, drop = FALSE], units = aw_units
--- a/R/plan_commodities_mean_reversion.R
+++ b/R/plan_commodities_mean_reversion.R
@@ -323,6 +323,8 @@ plan_commodities_mean_reversion <- function() {
     # n_months is a count. avg_dd_duration/max_dd_duration are drawdown
     # lengths measured in the return series' own periodicity (months, for
     # CMR's monthly returns) — classified as "count" (periods), not "days".
+    # ann_rf (#677 slice 4, #691) is a decimal fraction, same convention as
+    # cagr (round(sr$ann_rf, 4) above, never *100).
     row <- cmr_summary[cmr_summary$lookback == p, , drop = FALSE]
     if (nrow(row) == 1L) {
       metric_cols <- setdiff(names(row), "lookback")
@@ -330,7 +332,8 @@ plan_commodities_mean_reversion <- function() {
       cmr_units <- c(
         n_months = "count", sharpe = "ratio", cagr = "fraction",
         vol = "fraction", max_dd = "fraction",
-        avg_dd_duration = "count", max_dd_duration = "count"
+        avg_dd_duration = "count", max_dd_duration = "count",
+        ann_rf = "fraction"
       )
       historicaldata::hd_metric_record(con, uu, wide, units = cmr_units)
     }
--- a/R/plan_drif.R
+++ b/R/plan_drif.R
@@ -612,14 +612,17 @@ plan_drif <- function() {
   # Record full-period metrics row
   # Units (#640): drif_metrics stores cagr/vol/max_dd/hit_rate/bench_* as
   # decimal fractions (see calc_metrics() above), sharpe/bench_sharpe are
-  # scale-free ratios, and months is a period count.
+  # scale-free ratios, and months is a period count. ann_rf (#677 slice 4,
+  # #691) is a decimal fraction, same convention as cagr
+  # (ann_rf <- mean(df$rf_ret, na.rm = TRUE) * 12 above, never *100).
   full_row <- drif_metrics[drif_metrics$period == "Full Period", , drop = FALSE]
   if (nrow(full_row) == 1L) {
     metric_cols <- setdiff(names(full_row), "period")
     drif_units <- c(
       months = "count", cagr = "fraction", vol = "fraction",
       sharpe = "ratio", max_dd = "fraction", hit_rate = "fraction",
-      bench_cagr = "fraction", bench_vol = "fraction", bench_sharpe = "ratio"
+      bench_cagr = "fraction", bench_vol = "fraction", bench_sharpe = "ratio",
+      ann_rf = "fraction"
     )
     historicaldata::hd_metric_record(
       con, uu, full_row[, metric_cols, drop = FALSE], units = drif_units
--- a/R/plan_factormax.R
+++ b/R/plan_factormax.R
@@ -508,14 +508,17 @@ plan_factormax <- function() {
   # Record full-period metrics row
   # Units (#640): fm_metrics stores cagr/vol/max_dd/hit_rate/bench_* as
   # decimal fractions (see .fm_metrics calc_metrics()), sharpe/bench_sharpe
-  # are scale-free ratios, and months is a period count.
+  # are scale-free ratios, and months is a period count. ann_rf (#677
+  # slice 4, #691) is a decimal fraction, same convention as cagr
+  # (ann_rf <- mean(df$rf_ret) * 12 above, never *100).
   full_row <- fm_metrics[fm_metrics$period == "Full Period", , drop = FALSE]
   if (nrow(full_row) == 1L) {
     metric_cols <- setdiff(names(full_row), "period")
     fm_units <- c(
       months = "count", cagr = "fraction", vol = "fraction",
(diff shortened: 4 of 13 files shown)


JohnGavin:
Reviewed. The unit assignments are right and the mom_prepeak correction is right — I checked each against the same two independent sources you did (the metrics-computing function, and `R/plan_leaderboard.R`'s `.norm_*()` conversions), and they agree.

The `pso_optimal_register_runs` finding is a good catch and worth stating plainly: it uses `intersect()` where the other ten use `setdiff()`, so it is the one writer immune to this defect class — by accident rather than by design, but immune. That asymmetry is itself worth carrying into #668.

## One thing to be honest about before this merges

The test guards the wrong direction.

`required` is a hardcoded list in the test file. So the test fails when someone
**removes** an entry from a units map — which is what you demonstrated — but it
passes when someone **adds** a column to a metrics target:

```
add ann_vol_downside to ltr_metrics
  → enters metric_cols automatically (setdiff)
  → not in ltr_units          → hd_metric_record() aborts
  → not in test's `required`  → test passes
  → error = "continue"        → tar_make() exits 0
```

That is #691, unchanged, with a green test suite on top of it.

Put differently: this test would not have caught #691. At the time `ann_rf` was
added, `required` would not have listed it either — the same hand-maintenance
that failed for the units map fails identically for `required`. We now have three
lists that must agree instead of two.

That is not a reason to hold the fix — six registry writers are erroring on `main`
right now and this repairs them. But the file's header says "Regression test for
#691" and a future reader will take that at face value. It needs to say, in the
header, exactly which direction it covers and which it does not.

Merging, with two follow-ups tracked:

1. **State the limitation in the test header**, and derive `required` from source
   where it is tractable — you already wrote the AST extractor; pointing it at
   each metrics target's `tibble(...)` call would make `required` self-updating
   for the eight plans that build their metrics inline. (`mom_prepeak`,
   `stk`/`xgb` compute theirs in a shared helper elsewhere; those may have to
   stay declared, but then the declaration is 3 lists shrinking to 1.)
2. **#693** — the reason this went unseen for four commits is not that a test was
   missing, it is that `error = "continue"` makes an errored target invisible and
   nothing runs `scripts/check_pipeline_errors.R`. A test that fails in CI is a
   different (and weaker) guard than a build that refuses to report success.

The `bt.metric` backfill for the mislabelled mom_prepeak rows stays open as you
noted.

Issue #680, opened 2026-08-18
scripts/verify.sh never touches the targets DAG — a change can pass full verification and leave main unbuildable


JohnGavin:
`scripts/verify.sh` is documented as *"the one command to run before claiming a change is verified"* (`.claude/CLAUDE.md`). It does not touch the targets DAG. A change can pass it completely and leave `main` unbuildable — that happened on 2026-08-17.

## What verify.sh actually runs

From its own header:

```
#   1. parse("_targets.R")                                  (always)
#   2. root testthat suite  (tests/testthat/)                (full mode only)
#   3. package testthat suite (packages/historicaldata/)      (full mode only)
```

That is the complete list. `parse("_targets.R")` checks the file is syntactically valid R — not that the pipeline definition is coherent, not that any target builds.

There is **no `tar_validate`, `tar_manifest`, `tar_network` or `tar_outdated`** anywhere in `scripts/` or `.github/workflows/` (grepped, zero hits). Nothing in the project validates the DAG at any point.

## The incident

PR #678 (`1167311`, #677 slice 1) added an rf-coverage guard to the `ltr_portfolio` target. It passed `scripts/verify.sh` in full:

```
::VERIFY:ROOT_SUMMARY:: total=468 failed=3 skipped=0
::VERIFY:PKG_SUMMARY::  total=694 failed=1 skipped=15
=== scripts/verify.sh: PASS ===
```

Both baselines matched exactly. It was merged. The first `tar_make` afterwards:

```
x ltr_portfolio errored
x ltr_metrics   errored — could not load dependency ltr_portfolio
x ltr_subperiod errored — could not load dependency ltr_portfolio
x leaderboard   errored — could not load dependency ltr_portfolio
✖ errored pipeline [3 completed, 195 skipped]
```

Four targets down, including `leaderboard`. Fixed in #679.

## Why this is worse than a missing check

**The exit code lied.** `tar_make` returned **0** despite four errored targets, because the project sets `tar_option_set(error = "continue")`. So even a naive `tar_make` added to `verify.sh` would not have caught this — it would have exited 0 and reported success.

[...]

PR #694, opened 2026-08-18
feat(scripts): build.sh (#693) + derive unit-map coverage from source (#691 follow-up)

### LINKED ISSUE #691: ann_rf added to 11 metrics targets and to none of the registry unit maps — every registry writer aborts, hidden by error = "continue"

### LINKED ISSUE #693: nothing runs check_pipeline_errors.R — building the pipeline and checking it built are two commands, and only one is habitual
Nothing runs `check_pipeline_errors.R` — building the pipeline and checking it built are still two commands, and only one of them is habitual

## The gap

#680 established that `tar_make()` exiting 0 proves nothing, because
`docs/_targets.R` sets `error = "continue"` project-wide. PR #690 added
`scripts/check_pipeline_errors.R` to close that — it reads
`tar_meta(fields = error)` directly instead of trusting the exit code.

Nothing invokes it.

- No GitHub workflow runs `tar_make()` at all (`data-poll.yml`, `flake-guard.yml`,
  `link-audit.yml`, `pkgctx-check.yml`, `pytest.yml`, `r-tests.yml` — none of them).
  The pipeline is only ever built locally, in the main checkout, by hand.
- `scripts/verify.sh` deliberately does not run it (a worktree has no store).
- So the check runs exactly when someone remembers to type a second command
  after the first one finishes — which, for a build that takes 40+ minutes, is
  the moment attention is furthest from it.

#691 is the demonstration: six registry writers errored on every build for four
commits. It was found on the first occasion anyone ran the new check.

## Fix

Same shape as #569's answer to "how do I verify a change": make it one command.

```
scripts/build.sh          # tar_make() then check_pipeline_errors.R
```

- Runs `targets::tar_make()` against `docs/_targets.R` / `docs/_targets`
  (explicit `script =` and `store =`, per #680 — never rely on the gitignored
  `_targets.yaml`).
- Then runs `scripts/check_pipeline_errors.R` unconditionally, **including when
  `tar_make()` itself failed** — a crashed build still leaves errored targets
  worth naming.
- Exits non-zero if either step reports a problem, so the failure is impossible
  to walk past.
- Prints the errored-target list last, so it is the final thing on screen rather
  than buried 900 lines up in the build log.

Then `.claude/CLAUDE.md`'s "Verifying a change" table gains a second row:
`scripts/verify.sh` for a change, `scripts/build.sh` for a build — and neither
is a substitute for the other (`verify.sh` runs no target body; `build.sh` runs
no test).

## Why not a CI job

Building this pipeline takes 40+ minutes locally with a warm nix shell and hits
live data endpoints. Putting it in CI is a separate and much larger decision
(#673 is already about the fact that `equity_daily` has no scheduled refresh).
This issue is only about making the local build self-checking.

## Related

- #680 — the parent finding: `verify.sh` never touched the DAG; also names the
  still-uncovered render-time `tar_read()` surface
- #691 — the six errored registry writers this would have surfaced on day one
- #569 — `scripts/verify.sh`, the precedent this copies

JohnGavin:
## Summary

Two related pieces of work from #693 and the #692 review comment.

### Part 1 — `scripts/build.sh` (#693)

Building the pipeline and checking it built were two commands, and only the
first was habitual — #691's six errored registry writers were only found
because someone ran `scripts/check_pipeline_errors.R` by hand after a build.

`scripts/build.sh`:

1. **Refuses to run outside the main checkout.** Detects a linked worktree
   via `git rev-parse --absolute-git-dir` containing `/.git/worktrees/` and
   exits 2 with a clear message. Verified empirically: running it from this
   PR's own worktree prints
   `Refusing to build from a linked worktree: <path>` and exits 2.
2. Pre-flight-checks that `nix develop` can actually provide a shell
   (`nix develop "$REPO_ROOT" --command true`) before attempting the real
   build, so a broken flake is reported as "did not run" (exit 2), not
   misattributed as "targets errored" (exit 1).
3. Runs `targets::tar_make(script = "docs/_targets.R", store = "docs/_targets")`
   — explicit `script=`/`store=`, never the gitignored `_targets.yaml`
   (#680).
4. Runs `scripts/check_pipeline_errors.R` **unconditionally**, including
   when `tar_make()` itself exited non-zero — a crashed build still leaves
   errored targets worth naming.
5. Prints a short final summary block **after** both steps, so the
   errored-target list (already printed by step 4, right above it) is near
   the end of the output rather than buried under a 40+ minute build log.

**Exit codes:** `0` clean build (tar_make() exit 0, no errored targets).
`1` build ran but reported a problem — either `check_pipeline_errors.R`
found errored targets, or `tar_make()` itself exited non-zero (even if the
store it left behind happens to read clean — a crash still means the store
may not reflect a complete build, so that case is never silently
upgraded to a pass). `2` did not run at all — wrong location/worktree, nix
shell unusable, or `check_pipeline_errors.R` couldn't read a store at all.

**What I ran vs. reasoned about:** I ran `bash -n` (syntax OK) and executed
the script for real from this worktree — it correctly detected the
worktree and exited 2 with the documented message (see transcript below). I
did **not** run the actual `tar_make()` / `check_pipeline_errors.R` path
(forbidden by task scope — a worktree must not build a store that could
race the main checkout's). That path is code-reviewed but not executed by
me; please run it for real in the main checkout.

```
$ scripts/build.sh
!!! Refusing to build from a linked worktree: /Users/.../worktrees/agent-a94289d94af3523b9
!!! git-dir: /Users/.../historical/.git/worktrees/agent-a94289d94af3523b9
!!! A worktree has no targets store of its own and must not build one that
!!! could race the main checkout's docs/_targets store. Run scripts/build.sh
!!! from the main checkout instead.
!!! BUILD DID NOT RUN. This is NOT a pass.
$ echo $?
2
```

### `check_pipeline_errors.R` — real defect found and fixed

`tar_meta(store = store_path, fields = error, complete_only = TRUE)` was
printing `Complete targets checked: 0` on a clean pipeline, directly above
`PASS: no errored targets` — reading as "nothing was inspected" rather than
"nothing was wrong".

**Empirical finding** (I could not touch the live `docs/_targets` store per
task scope, so I verified against `targets::tar_meta`'s own installed
source in this nix shell, printed via `print(targets::tar_meta)`):

```r
out <- out[, base::union("name", fields), drop = FALSE]
if (complete_only) out <- out[stats::complete.cases(out), , drop = FALSE]
```

With `fields = error`, `out` has exactly two columns: `name`, `error`.
`complete.cases()` on that requires **both** non-NA — so `complete_only =
TRUE` already filters down to the errored set before the script's own
`errored <- meta[!is.na(meta$error), ]` line even runs. That line was a
no-op on an already-filtered result, and `nrow(meta)` was the errored
count, not the inspected count.

Fixed: dropped `complete_only = TRUE`, added `targets_only = TRUE` (the
default also counts functions/other global objects, not just targets — so
the count would never have meant "targets" even after fixing
`complete_only`), and now prints `Targets inspected: N` / `Targets errored:
M` as two separate, honest numbers. Full reasoning is inline in the script.

Also added a row to `.claude/CLAUDE.md`'s "Verifying a change" table:
`verify.sh` and `build.sh` are explicitly **not** substitutes for each
other — `verify.sh` runs no target body, `build.sh` runs no test.

### Part 2 — the #691 regression test's limitation (#692 review)

The review on #692 stated the problem precisely: `required` in
`test-registry-unit-map-coverage.R` was a hardcoded list in the test file,
so the test failed when a units map **removed** a covered column but
**passed** when a metrics target **added** one — which is exactly the shape
of #691's actual defect. "This test would not have caught #691."

**(a) Header wording — done.** The file's header now states plainly which
direction the (now largely fixed, see below) test covers and which
residual cases still rely on hand-maintenance, so a future reader can't
take a green run as proof #691 cannot recur.

**(b) Derive `required` from source — done for 9 of 11 cases.**

Two new AST extractors, alongside the existing `.extract_units_map()`:

- `.derive_required_from_tibble(file, fn_name, exclude)` — finds the
  `fn_name <- function(...) {...}` assignment, walks its body for
  `tibble()`/`tibble::tibble()`/`dplyr::tibble()` call(s), and returns the
  union of their named-argument names minus a small stable `exclude` set
  (the function's own known character/Date columns — auto-skipped by
  `hd_metric_record()`'s numeric-only path regardless of unit coverage).
  Used for **8 cases**: `ltr`, `aw`, `tom`, `rsc`, `cmr`, `fm`, `drif`,
  and `stk`/`xgb` share one function (`calc_backtest_metrics()` in
  `plan_stock_backtest.R`) — **9 units maps** from 8 function walks.
- `.extract_intersect_allowlist(file, varname)` — for `pso_optimal`, which
  uses an explicit `intersect(c(...), names(full_row))` allow-list rather
  than a `setdiff()`-based exclusion (unlike every other writer). Reads
  that literal allow-list directly, so the writer and the test can never
  drift apart.

**1 of 11 stays hand-declared** — `mom_prepeak_units`. Tried deriving it;
stopped because it needs strictly more than the other cases:
`.mom_prepeak_compute_metrics()`'s `tibble()` lives in
`packages/historicaldata/R/utils_mom_prepeak_metrics.R`, a **different
file** from `R/plan_mom_prepeak.R` (which owns the unit map), and `ann_rf`
is appended **after** that `tibble()` returns via a `$<-` mutation
(`m$ann_rf <- round(.mom_prepeak_ann_rf(...) * 100, 2)`), which no
AST walk over a single `tibble()` call can see. Deriving this one case
would need a cross-file walk plus `$<-`-mutation tracking — meaningfully
more machinery than the single-function walk that handles the other 9,
for one case. Declared, with the reasoning in the case's own comment.

**Both failure directions demonstrated, then reverted** (see the two
Edit/revert pairs and the test-run transcript below — no diff remains in
either source file):

1. **Deletion** (the original, already-caught direction): removed `ann_rf`
   from `ltr_units` in `plan_ltr_momentum.R` → test failed:
   `ltr_units (plan_ltr_momentum.R) is missing unit declarations for: ann_rf`.
2. **Addition** (the direction #691 actually needed, now newly caught):
   added a `dummy_demo_col = 1` argument to `plan_factormax.R`'s
   `calc_metrics()` tibble → test failed:
   `fm_units (plan_factormax.R) is missing unit declarations for: dummy_demo_col`.

```
══ Failed ══════════════════════════════════════════════════════════════════
── 1. Failure ('test-registry-unit-map-coverage.R:372:9'): ltr_units covers its
Expected `missing` to equal `character(0)`.
Differences:
`actual`:   "ann_rf"
`expected`:
ltr_units (plan_ltr_momentum.R) is missing unit declarations for: ann_rf

── 2. Failure ('test-registry-unit-map-coverage.R:372:9'): fm_units covers its m
Expected `missing` to equal `character(0)`.
Differences:
`actual`:   "dummy_demo_col"
`expected`:
fm_units (plan_factormax.R) is missing unit declarations for: dummy_demo_col
```

Both edits were reverted immediately after this run; `git diff --stat` on
both files shows no residual changes.

## `scripts/verify.sh` output (full run, foreground)

```
PASS: _targets.R parses
PASS: docs/_targets.R parses
PASS: docs/_targets.R is structurally valid (tar_validate, 7.42s)

SKIP count this run: 0
PASS: failures match the known baseline exactly (test-crypto-momentum.R,
test-qa-summary-deps.R, test-stock-backtest.R — all pre-existing per #569)

SKIP count this run: 15
PASS: failures match the known baseline exactly (test-mom-prepeak-portfolio.R)

```
Root total=587, package total=695 (13 test_that blocks in the modified file,
same count as before — only what `required` derives from changed).

## Test plan

- [x] `scripts/verify.sh` (full, foreground) — PASS, both baselines exact
- [x] `bash -n scripts/build.sh` — syntax OK
- [x] `scripts/build.sh` executed for real from this worktree — correctly
      refuses with exit 2 and the documented message
- [ ] `scripts/build.sh`'s `tar_make()` + `check_pipeline_errors.R` path —
      **not run by me** (forbidden in a worktree); please run in the main
      checkout
- [x] Deletion-direction regression demo (ltr_units/ann_rf) — fails as
      expected, reverted
- [x] Addition-direction regression demo (fm_units/dummy_demo_col) — fails
      as expected, reverted
- [x] `targets::tar_meta()` `complete_only`/`targets_only` semantics
      verified against the installed package's own source

Closes #693.

Does not close #691 — the structural fix (#668) and the `bt.metric`
backfill for the mislabelled `mom_prepeak` rows remain open, as noted in
the #692 review.


JohnGavin:
Reviewed. `build.sh` is right, the `complete_only` diagnosis is right, and the
derivation is real rather than decorative — I checked that last claim rather than
taking it, by evaluating the helpers and printing what each case actually derives:

```
ltr_units          mode=tibble    n=10  months,cagr,vol,max_dd,sharpe,ann_rf,hac_sharpe,hac_tstat,avg_long,avg_short
aw_units           mode=tibble    n= 7  years,n_days,cagr,vol,max_dd,sharpe,ann_rf
tom_units          mode=tibble    n=13  n_days,years,cagr_tom,...,ann_rf_tom,...,pct_in_tom
rsc_units          mode=tibble    n= 7  cagr,vol,sharpe,ann_rf,max_dd,hac_tstat,hac_sharpe
cmr_units          mode=tibble    n= 8  n_months,sharpe,cagr,vol,ann_rf,max_dd,avg_dd_duration,max_dd_duration
fm_units           mode=tibble    n=10  months,cagr,vol,sharpe,ann_rf,max_dd,hit_rate,bench_*
drif_units         mode=tibble    n=10  months,cagr,vol,sharpe,ann_rf,max_dd,hit_rate,bench_*
stk_units          mode=tibble    n= 8  months,cagr,vol,sharpe,ann_rf,max_dd,avg_long,avg_short
xgb_units          mode=tibble    n= 8  months,cagr,vol,sharpe,ann_rf,max_dd,avg_long,avg_short
pso_units          mode=intersect n= 5  months,opt_cagr,opt_vol,opt_sharpe,opt_maxdd
mom_prepeak_units  mode=declared  n=10  n_months,sharpe,cagr,vol,max_dd,...,ann_rf
```

Every derived set reproduces the hardcoded list it replaced, exactly, and every
one carries `ann_rf` except `pso_units` — which correctly does not, for the
`intersect()` reason you documented. Ten of eleven derived, one declared with a
reason I accept: cross-file plus a `$<-` mutation is more machinery than one case
justifies.

Deriving `pso_units` from `pso_cols` is a nice touch — the extractor and the
writer read the same literal, so they cannot drift.

## One hole left, and it is the same shape as the bug

`.extract_tibble_column_names()` returns `character(0)` when it finds no
`tibble()` call. `.derive_required_from_tibble()` passes that straight through,
`setdiff(required, names(units_map))` is then empty, and the test **passes**.

So a refactor from `tibble(...)` to `data.frame(...)`, `as_tibble(...)`, or a
`transmute()` pipeline in any of those nine functions silently empties `required`
and the case goes green while checking nothing.

`.extract_function_body()` and `.extract_intersect_allowlist()` both `cli_abort()`
when their target is missing — correctly. The tibble-column walk is the one step
that returns an empty result instead of failing, which is `fail-loud-not-null`'s
exact prohibition inside a test written to enforce it.

Three lines fixes it: for `mode = "tibble"` and `mode = "intersect"`, abort if
`length(required) == 0` naming the function and file, since no metrics
constructor in this repo legitimately emits zero numeric columns. Tracking as a
follow-up rather than holding the PR — `build.sh` matters more than the residual
hole, and the derivation as shipped is a real improvement over the hardcoded list.

## Verified

My own `scripts/verify.sh` on this branch, plus the extractor evaluation above.
I have not executed `build.sh`'s `tar_make()` path either — I will run it for
real in the main checkout after this merges, which is also the first end-to-end
exercise of the exit-code logic.

Issue #697, opened 2026-08-20
7 dashboards have no build_info() footer -- degrades #695 Check 3 to git-date fallback


JohnGavin:
Seven dashboards call no `build_info()` and so carry no `Built` timestamp
footer in their rendered `.html`:

- `avoid-worst-days.qmd`
- `bdbb-sol.qmd`
- `drif.qmd`
- `factor-max.qmd`
- `jst-dashboard.qmd`
- `macro-defense-rotation.qmd`
- `negative-results.qmd`

Confirmed via `grep -rLE "build_info\(" docs/*.qmd` against the current repo
(2026-08-20). Four of these (`drif`, `factor-max`, `jst-dashboard`,
`negative-results`) are genuinely-empty redirect stubs (0 R chunk fences,
`> This page has moved.` + a `<meta http-equiv="refresh">`), so a missing
footer there is expected and low-priority. The other three --
`avoid-worst-days.qmd`, `bdbb-sol.qmd`, and `macro-defense-rotation.qmd` --
are real content pages (`macro-defense-rotation.qmd` alone has 21 R chunks)
that render real output with no footer at all.

## Why this matters

Per `vignette-build-info-block`, every rendered dashboard is supposed to
carry a `Built` timestamp footer via `build_info()`
(`docs/vignette_utils.R:205`). `scripts/check_dashboard_freshness.R` (#695)
Check 3 (data staleness -- page render time vs. the newest `tar_meta()`
build time among the targets that page reads) reads that footer as its
primary render-time source, and falls back to the `.html`'s own git commit
date when no footer is found -- a documented, explicit fallback, not a
silent one. For these seven pages, Check 3 always uses the fallback, which
is coarser: a git commit can predate or postdate the actual `quarto render`
that produced the committed content (e.g. a `git commit --amend`, or several
renders batched into one commit), so the reported staleness gap for these
pages is only as accurate as the commit history around them.

## What to do

Add a `build_info()`-driven footer to the three real content pages
(`avoid-worst-days.qmd`, `bdbb-sol.qmd`, `macro-defense-rotation.qmd`),
following the pattern already used by the 8 pages that have one (e.g.
`leaderboard.qmd`, `stock-backtest.qmd`). The four redirect stubs can either
gain a trivial footer too (for consistency) or stay as-is with a one-line
note in the page explaining why -- worth a small discussion rather than
assuming either way.

Not fixed in #695's PR -- that PR is detection-only (Option A) and
[...]

Issue #695, opened 2026-08-18
published site is stale: 8 of 15 dashboards have HTML older than their .qmd, and the live leaderboard is missing Holdout (#666) and Financing (#689)


JohnGavin:
The published site is stale: 8 of 15 dashboards have committed HTML older than their own `.qmd`, and the live leaderboard is missing two merged features

## Verified against the live site, not inferred

GitHub Pages serves this repo from `branch: main, path: /docs`, legacy build type
— confirmed via `gh api repos/JohnGavin/historical/pages`. No workflow renders
anything (`data-poll`, `flake-guard`, `link-audit`, `pkgctx-check`, `pytest`,
`r-tests` — none of them run `quarto render`). **The committed `.html` files
*are* the published site.** A merged `.qmd` change reaches readers only when
somebody re-renders by hand and commits the output.

`curl https://johngavin.github.io/historical/leaderboard.html`:

| string | live site | freshly rendered |
|---|---:|---:|
| `Holdout` | **0** | 2 |
| `Financing` | **0** | 2 |

So the live leaderboard shows neither the **Holdout partition** (#666, merged) nor
the **Financing Sensitivity** tab (#689, merged). Both are on `main`. Neither is
visible to anyone reading the site.

## It is not one dashboard

Comparing each dashboard's last `.qmd` commit against its last `.html` commit:

| dashboard | `.qmd` | `.html` | gap |
|---|---|---|---:|
| leaderboard | 2026-08-18 | 2026-08-05 | 13d |
| stock-backtest | 2026-08-05 | 2026-07-01 | 35d |
| bdbb-sol | 2026-08-04 | 2026-07-01 | 34d |
| drif | 2026-08-04 | 2026-07-01 | 34d |
| factor-max | 2026-08-04 | 2026-07-01 | 34d |
| jst-dashboard | 2026-08-04 | 2026-07-01 | 34d |
| negative-results | 2026-08-04 | 2026-07-01 | 34d |
| falsification | 2026-08-04 | 2026-07-03 | 32d |

Eight of fifteen. And source-vs-render is the *lower* bound on staleness: every
one of these renders `tar_read()` chunks, so even a dashboard whose `.qmd` has not
moved is serving **data** from whenever it was last rendered. The pipeline has been
rebuilt many times since 2026-07-01.

## Why this went unseen

This is the third surface #680 named, and the only one still uncovered:

| surface | caught by | status |
|---|---|---|
[...]

PR #698, opened 2026-08-20
feat(scripts): detect dead tar_read() refs and dashboard staleness (#695)

JohnGavin:
## Summary

Implements #695's "Option A" (detection only): `scripts/check_dashboard_freshness.R`, wired into `scripts/verify.sh` (Check 1 + Check 2) and `scripts/build.sh` (adds Check 3). Closes the third render-time surface #680 named and #695's issue body documented as the only one still uncovered — a `docs/*.qmd` chunk calling `tar_read()` for a target that no longer exists fails only at an actual `quarto render`, and a committed `.html` older than its own `.qmd` (or older than the data it reads) silently keeps serving stale content since GitHub Pages serves the committed `.html` directly.

**Re-rendering the 8 currently-stale dashboards is explicitly out of scope** — that is #695's other half, tracked by the issue itself, not this PR.

## Provenance — this builds on a prior session's uncommitted work

A previous agent dispatched against this same task got most of the way (the script, the wiring into both scripts, the test file, and the snapshot) and was killed mid-task before committing. I recovered its files from a salvage location and treated them as a **starting point, not a finished product** — I own everything in this diff and re-verified all of it myself rather than trusting the prior report. Concretely, what I did on top of the recovered files:

- Diffed `scripts/verify.sh` and `scripts/build.sh` in full against the current worktree versions (not just trusted "these are full files from `012c78d`") — confirmed the only changes in both are the freshness wiring plus doc-comment updates; no unrelated hunks. Diffs quoted in a review comment below if useful, omitted here for length.
- Read all 679 lines of `check_dashboard_freshness.R` and all 378 of the test file end to end.
- Independently re-ran and re-verified every empirical claim the script's own header comment makes (see "What I verified" below) rather than accepting the comments as given.
- Found and corrected one factual error in the recovered work: its planned follow-up issue was going to say "six dashboards have no `build_info()` footer." I checked and it's **seven** — `macro-defense-rotation.qmd` (21 R chunks, a real content page, not a redirect stub) also has none. Filed the corrected issue as #697 (see below), not six.
- Actually ran `scripts/verify.sh` full mode to completion (previous attempts had reportedly stalled/been killed) and confirmed PASS with baseline-matching failures.
- Performed the "Check 1 actually fires" demonstration below, which had not been done.

## Extraction approach

`knitr::purl(documentation = 0, quiet = TRUE)` tangles each `.qmd`'s R chunks to a temp `.R` file, which is `parse()`'d (never evaluated) into an expression list, then walked for calls headed by `tar_read`, `tar_read_raw`, or `safe_tar_read` (bare or `pkg::fn` form), taking the first argument. **Not a regex over the `.qmd` text** — this repo already spent a whole session (#691→#696) on the cost of fragile regex-based extraction, and this script deliberately avoids repeating that.

I independently re-confirmed the extraction script's own claims against the current repo rather than trusting its comments:

- `grep -rnE "tar_read(_raw)?\(([^\"'a-zA-Z_.]|[a-zA-Z_.]+\()" docs/*.qmd"` → **no matches**, confirming every `tar_read()`/`tar_read_raw()` call across all 15 dashboards uses a string-literal or bare-symbol first argument — nothing the extractor can't statically resolve.
- The 4 files claimed to `purl()` to zero expressions (`drif.qmd`, `factor-max.qmd`, `jst-dashboard.qmd`, `negative-results.qmd`) each have `grep -c '\`\`\`{r' <file>` == 0 — genuinely-empty redirect stubs, not a broken extractor swallowing real chunks.
- `falsification.qmd`'s two inline-expression target references (`cg_dag` at line 806, `cg_caption` at line 986, both via `` `r { ... } ` `` inline syntax) are real and are **not** duplicated in any fenced chunk in that file — confirmed by grep. `knitr::purl()` does not tangle inline expressions (by design, no supported non-evaluating API for it), so this is a genuine, documented coverage gap, not an oversight. It affects two target names in one file; Check 1 cannot catch a dead reference to either, and (once Check 3 runs) those two targets are excluded from that page's own-targets staleness bound.

## An unresolvable reference aborts, never silently skips

Per `.claude/rules/fail-loud-not-null.md`, `.cdf_extract_read_targets()` calls `cli::cli_abort()` — not a silent skip — the moment a matched `tar_read`-family call's first argument is neither a string literal nor a bare symbol, and also aborts on a zero-argument call. Both paths have `expect_snapshot(error = TRUE, ...)` coverage. A page yielding **zero** target references (e.g. a redirect stub, or `index.qmd`, which purls to 27 expressions but references no targets) does **not** silently pass as "fresh" by omission — it is a legitimate zero-length entry in the per-page target map, distinguished from a broken extraction via `.cdf_has_r_chunk_fence()`: the script aborts only when `purl()` finds zero expressions **and** the raw source has at least one fenced R chunk (meaning `purl()` silently dropped something real). I re-verified this distinction is exercised by both directions of the test suite (`redirect stub, zero R chunks` passes cleanly; `chunk fence present but purl() extracts zero expressions` aborts).

## Three checks, where each is wired, and the exit-code contract

| Check | What | Needs a store? | Wired into | Failure mode |
|---|---|---|---|---|
| 1 — dead target references | `tar_read()`-family calls vs. `tar_manifest()` | No | `verify.sh` full mode | **Always** a hard failure (exit 1) |
| 2 — source-vs-render staleness | `.qmd` git commit date vs. `.html`'s | No | `verify.sh` full mode | Informational by default; `HD_FAIL_ON_STALE_DASHBOARDS=1` escalates |
| 3 — data staleness | page's own referenced targets' newest `tar_meta()` time vs. page render time (`build_info()` footer, else git-date fallback, labelled) | **Yes** | `scripts/build.sh` Step 3 (main checkout only) | Same escalation flag as Check 2 |

Check 1 is unconditional because a dead reference **will** break the next render — there's no legitimate reason to tolerate it. Checks 2/3 default to informational because #695's own evidence is that 8 of 15 dashboards are stale *right now* — defaulting staleness to a hard failure would turn `verify.sh`/`build.sh` red on `main` today, before the separate re-render PR lands. `HD_FAIL_ON_STALE_DASHBOARDS=1` is the documented, explicit escalation path once that PR ships.

Check 3 isn't in `verify.sh` because it needs a real `docs/_targets` store, which a worktree doesn't have and must not build (would race the main checkout's store — see `.claude/CLAUDE.md`). It's wired into `build.sh` Step 3 instead, running unconditionally alongside Step 2 (`check_pipeline_errors.R`) even after a `tar_make()` failure, for the same reason Step 2 does: `error = "continue"` means a crashed/partial build still leaves a store worth inspecting. `build.sh`'s exit-code contract (0/1/2) mirrors `check_pipeline_errors.R`'s existing pattern.

**I could not run Check 3** — no store in a worktree. I read all of Step 3's wiring in `build.sh` and reasoned about it (see "What I executed vs. reasoned about" below); I did not execute it.

## Timing

Measured directly (warm nix shell, `docs/_targets.R` at current `main`, 768 targets):

- Extraction (`knitr::purl()` across all 15 `.qmd` files): **0.24–0.35s**
- `targets::tar_manifest(docs/_targets.R)`: **6.51–7.24s**
- Check 1 + Check 2 combined added cost to `verify.sh` full mode: **~7–8s**
- `verify.sh --quick`: **unaffected** — confirmed by running it after this change; the freshness fragment is only spliced into the full-mode R script, `--quick`'s R script doesn't reference it at all, and the `--quick` output contains no mention of `check_dashboard_freshness.R`.

**My argument on whether 7–8s belongs in `verify.sh` full mode:** yes. `verify.sh` full mode already pays a near-identical cost for `tar_validate()` (6.7–6.9s, same order of magnitude, evaluating the same `docs/_targets.R`) as part of its existing structural checks, and full mode is dominated by the two `testthat` suites (root + package, several minutes total) — an extra 7–8s is a low-single-digit percentage addition to an already multi-minute run. `--quick` is the fast path for iterative work and stays exactly as fast as before this PR (verified above), which is where the 5-second-class budget actually matters. Check 1 could in principle share `tar_manifest()`'s cost with `tar_validate()`'s call, but `tar_validate()` returns `NULL` invisibly (confirmed by inspection) — there's no cheaper way to get the current target-name set, so paying the manifest cost twice in full mode (once for validate, once for manifest) is the real cost of this design, not a wasted one. I don't think Check 1 belongs anywhere else: it needs the current target list, which only exists after evaluating `docs/_targets.R`, and `verify.sh` full mode is the only place that already pays that price.

## Check 1 demonstration — it actually fires

Temporarily added a chunk `` `.demo_dead_ref_695 <- tar_read("no_such_target_xyz")` `` to `docs/index.qmd`, reran the script standalone:

```
=== Check 1: dead target references ===
FAIL: dead target references found (these WILL fail at render/tar_read time):
  [DEAD-REF] index.qmd -- references non-existent target(s): no_such_target_xyz
...
=== scripts/check_dashboard_freshness.R: FAIL ===
::VERIFY:DASHBOARD_FRESHNESS_STATUS:: 1
```

Reverted the chunk. `git diff --stat docs/index.qmd` after revert: **no output** (clean).

## What I executed vs. reasoned about

**Executed:**
- `check_dashboard_freshness.R` standalone (Check 1 + Check 2), against the real `docs/` tree, twice (clean, then with the injected dead reference above).
- Full `scripts/verify.sh` (foreground, to completion) — see output below.
- `scripts/verify.sh --quick` — confirmed unaffected timing/behaviour.
- All grep-based re-verifications of the script's own header-comment claims (extraction literal-only args, redirect-stub chunk counts, inline-expression coverage gap, footer-file count).

**Reasoned about, not executed:**
- `scripts/build.sh` Step 3 (`check_dashboard_freshness.R --data-staleness`) — needs a real `docs/_targets` store, which this worktree does not have and must not create. I read the wiring (capture/summary/exit-code handling) end to end and it mirrors the existing Step 2 pattern closely enough that I'm confident in it, but it has not actually run against a live store. Please run it in the main checkout before/after merge.

## `scripts/verify.sh` output (full mode, this branch, this commit)

```
=== scripts/verify.sh ===
Mode: full (parse + docs/_targets.R structural validate + root suite + package suite)
...
PASS: _targets.R parses
PASS: docs/_targets.R parses
PASS: docs/_targets.R is structurally valid (tar_validate, 6.72s)
PASS: dashboard freshness (scripts/check_dashboard_freshness.R Check 1 + Check 2)

SKIP count this run: 0
PASS: failures match the known baseline exactly (3/3, unchanged)

SKIP count this run: 15
PASS: failures match the known baseline exactly (1/1, unchanged)

```

Root suite: `total=625 failed=3 skipped=0` (up from the `012c78d` baseline of `total=590 failed=3 skipped=0` — the +35 is exactly the 35 new `test_that()` blocks in `tests/testthat/test-dashboard-freshness.R`, all passing). Package suite: `total=695 failed=1 skipped=15`, unchanged from baseline (this PR touches no package code). Both suites' failure **signatures** match baseline exactly — no regressions.

## Follow-up issue filed

#697 — 7 dashboards (not 6, correcting the recovered draft) have no `build_info()` footer, which degrades Check 3 to the git-commit-date fallback for exactly those pages. Not fixed here; #695's PR is detection-only and out of scope for touching dashboard content.

## Test plan

- [x] `scripts/verify.sh` full mode — PASS, baselines unchanged (see output above)
- [x] `scripts/verify.sh --quick` — PASS, unaffected timing
- [x] `parse("_targets.R")` and `parse("docs/_targets.R")` — both OK (part of verify.sh)
- [x] New tests (`tests/testthat/test-dashboard-freshness.R`, 35 `test_that` blocks) — all pass, part of the root suite run above
- [x] Manual Check 1 fire-and-revert demonstration — see above
- [ ] `scripts/build.sh` Step 3 against a real store — **not run**, needs the main checkout (see "What I executed vs. reasoned about")

Refs #695

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -20,6 +20,14 @@
 # the demonstration: six registry writers errored on every build for four
 # commits, surfaced only on the first occasion anyone ran the check by hand.
 #
+# STEP 3 (#695): after the store exists, this script also runs
+# scripts/check_dashboard_freshness.R --data-staleness, which needs a real
+# store to compare each docs/*.qmd page own referenced targets build times
+# against that page rendered .html -- the ONE surface of #695's three that
+# verify.sh cannot cover (verify.sh runs Check 1 + Check 2 of the same
+# script, which need no store -- see that script header comment for the
+# full three-check design).
+#
 # MAIN CHECKOUT ONLY. A worktree has no targets store of its own and must
 # not build one that could race the main checkout's docs/_targets store (see
 # .claude/CLAUDE.md "Verifying a change" and the worktree-location rule).
@@ -30,19 +38,23 @@
 #   scripts/build.sh
 #
 # Exit codes:
-#   0  tar_make() ran and scripts/check_pipeline_errors.R found no errored
-#      targets.
+#   0  tar_make() ran, scripts/check_pipeline_errors.R found no errored
+#      targets, and scripts/check_dashboard_freshness.R --data-staleness
+#      found no dead target references (and no staleness escalated to a
+#      failure -- see that script header for HD_FAIL_ON_STALE_DASHBOARDS).
 #   1  the build RAN but reported a problem: one or more targets errored
-#      (per check_pipeline_errors.R), and/or tar_make() itself exited
-#      non-zero (a crashed build still leaves a store worth inspecting, so
-#      check_pipeline_errors.R always runs -- see Step 2 below -- but a
+#      (per check_pipeline_errors.R), a dead target reference or escalated
+#      staleness (per check_dashboard_freshness.R), and/or tar_make() itself
+#      exited non-zero (a crashed build still leaves a store worth
+#      inspecting, so Steps 2 and 3 always run -- see below -- but a
 #      non-zero tar_make() exit is never silently treated as a pass even
 #      when the store it left behind happens to read clean).
 #   2  the build did NOT run at all: wrong location (not the main checkout,
 #      or run from a linked worktree), the nix shell itself could not be
-#      entered (pre-flight check below), or check_pipeline_errors.R could
-#      not read a store at all (no store / tar_meta() failed). This is NOT
-#      a pass -- never treat it as one.
+#      entered (pre-flight check below), check_pipeline_errors.R could not
+#      read a store at all (no store / tar_meta() failed), or
+#      check_dashboard_freshness.R --data-staleness could not run its checks
+#      at all. This is NOT a pass -- never treat it as one.
 
 set -euo pipefail
 
@@ -137,14 +149,37 @@ echo "--- check_pipeline_errors.R exited $CHECK_STATUS ---"
 echo ""
 
 # ---------------------------------------------------------------------------
+# Step 3: scripts/check_dashboard_freshness.R --data-staleness (#695). Runs
+# UNCONDITIONALLY, same reasoning as Step 2: a crashed or partially-errored
+# tar_make() (docs/_targets.R sets error = "continue" project-wide) still
+# leaves a store with real build times for whatever DID succeed, worth
+# checking. This re-runs Check 1 (dead target references) and Check 2
+# (source-vs-render staleness) as well as Check 3 (data staleness) -- see
+# that script header comment for the full three-check design and why Check
+# 3 specifically needs a real store (hence living here, not in
+# scripts/verify.sh). Staleness (Check 2/Check 3) is informational-only by
+# default (HD_FAIL_ON_STALE_DASHBOARDS=1 escalates it) -- dead references
+# (Check 1) are always a hard failure. See that script header for the full
+# exit-code contract; the same 0/1/2 meaning is reused here.
+# ---------------------------------------------------------------------------
+echo "--- Step 3: scripts/check_dashboard_freshness.R --data-staleness (#695) ---"
+set +e
+nix develop "$REPO_ROOT" --command Rscript scripts/check_dashboard_freshness.R --data-staleness
+DASHBOARD_STATUS=$?
+set -e
+echo "--- check_dashboard_freshness.R --data-staleness exited $DASHBOARD_STATUS ---"
+echo ""
+
+# ---------------------------------------------------------------------------
 # Final verdict -- printed LAST and kept short, so it is the final thing on
 # screen rather than buried 900+ lines up in the tar_make() build log
-# (#693). The errored-target list itself was already printed by Step 2,
-# directly above this block.
+# (#693). The errored-target/dead-reference/staleness lists were already
+# printed by Step 2 and Step 3, directly above this block.
 # ---------------------------------------------------------------------------
 echo "=== scripts/build.sh: summary ==="
-echo "tar_make() exit code:               $TAR_MAKE_STATUS"
-echo "check_pipeline_errors.R exit code:  $CHECK_STATUS"
+echo "tar_make() exit code:                              $TAR_MAKE_STATUS"
+echo "check_pipeline_errors.R exit code:                 $CHECK_STATUS"
+echo "check_dashboard_freshness.R --data-staleness exit: $DASHBOARD_STATUS"
 echo ""
 
 if [[ "$CHECK_STATUS" -eq 2 ]]; then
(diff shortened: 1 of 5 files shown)


JohnGavin:
Reviewed. The header's KNOWN LIMITATIONS section is the best part of this PR —
it states the extractor's coverage gap in measured numbers rather than hedging.
I re-derived the central claim rather than taking it:

```
falsification.qmd: purl() extracts 32 distinct target names from fenced chunks
                   7 distinct names appear in inline `r ...` expressions
                   inline-ONLY: cg_caption, cg_dag
```

Exactly the two the header claims. The `stock-backtest.qmd` finding holds too —
19 inline calls, every name also present in a fenced chunk, so zero net loss
there. Disclosed accurately and quantified precisely; that is the right way to
ship a partial extractor.

`.cdf_has_r_chunk_fence()` addresses the thing I most wanted guarded: a page that
purls to zero expressions aborts unless it genuinely has no R chunks, so the
"silently fresh because we extracted nothing" vacuous pass cannot happen. Good.

Warn-by-default with `HD_FAIL_ON_STALE_DASHBOARDS=1` is the right call and the
escalation is a real switch rather than an intention. Eight stale pages should
not turn `main` red before #695's other half exists.

## One gap the header does not cover: Quarto includes

`knitr::purl()` tangles only the fenced chunks physically present in the file it
is handed. It does **not** resolve `{{< include ... >}}`. So a `tar_read()` inside
an included `.qmd` is invisible to the extractor — silently, for Check 1 and
Check 3 both.

Measured, not assumed:

- `docs/_includes/` holds exactly one file, `build-info-footer.qmd`
- it contains zero `tar_read` calls
- exactly one page (`macro-defense-rotation.qmd:710`) uses an include at all

So nothing is missed today. But nothing would detect it if that changed, and the
failure is silent in the direction that matters — a target removed from the
pipeline while an included chunk still reads it, which is precisely what Check 1
exists to catch.

This belongs in KNOWN LIMITATIONS next to the inline-expression gap. Better still
as a cheap guard: if a `.qmd` contains `{{< include`, resolve the referenced file
and either extract from it too, or abort naming it. Tracking as follow-up rather
than holding the PR — the checks are a large net gain over nothing, and today the
gap is provably empty.

## Correction to something I told you

I said six dashboards lack a `build_info()` footer. Your #697 says seven, adding
`macro-defense-rotation`. **Six is right**, and my reasoning for it was wrong.

`macro-defense-rotation.qmd:710` gets its footer through
`{{< include _includes/build-info-footer.qmd >}}`, so `grep -LE "build_info\("`
misses it. The rendered page settles it: `macro-defense-rotation.html` contains
`Built</strong> 2026-06-16`.

Both counts came from one grep and neither was a proof — mine was too loose and
caught the include by accident, yours was too narrow and assumed a direct call is
the only route. The test that actually answers the question is
`grep -LE "build_info\(|_includes/build-info-footer" docs/*.qmd`, which returns
six, cross-checked against which `.html` files carry a `Built` string.

I have posted the correction on #697; strike `macro-defense-rotation` there. The
rest of that issue stands.

## Verified

My own `scripts/verify.sh` on this branch, plus the two derivations above. I have
**not** run Check 3 — a worktree has no store, as you said. I will exercise it in
the main checkout after this merges; that is the first end-to-end run of the
per-page data-staleness comparison.

PR #700, opened 2026-08-20
fix(scripts): resolve Quarto {{< include >}} in dashboard-freshness extractor

JohnGavin:
## Summary

`scripts/check_dashboard_freshness.R`'s extractor uses `knitr::purl()` to tangle each `docs/*.qmd` file and walk the resulting expressions for `tar_read()`/`tar_read_raw()`/`safe_tar_read()` calls. `knitr::purl()` only tangles the fenced chunks physically present in the file it is handed — it does not resolve Quarto's `{{< include path.qmd >}}` shortcode (that expansion happens inside `quarto render`'s own pandoc preprocessing, never inside knitr). So a `tar_read()` living only inside an included file was silently invisible to both Check 1 (dead references) and Check 3 (data staleness), with no error and no log line.

## Verifying the gap before fixing it

Confirmed against the real repo (2026-08-20):

- `docs/_includes/` contains exactly one file, `build-info-footer.qmd`.
- It contains zero `tar_read` calls.
- Exactly one page uses an include at all: `docs/macro-defense-rotation.qmd:710`.

So the gap was provably empty right now — the defect was the silence, not a currently-live bug. Confirmed the mechanism empirically with a synthetic fixture: `knitr::purl()` on a `parent.qmd` containing only `{{< include child.qmd >}}` (where `child.qmd` has one fenced `tar_read()` chunk) tangles to a completely empty file (zero expressions). `.cdf_has_r_chunk_fence(parent.qmd)` is `FALSE` (the fence lives in `child.qmd`, not `parent.qmd`), so the old zero-expressions-but-no-fence branch treated it exactly like `index.qmd`'s legitimate "no code here" case and never flagged the hidden reference.

## Design chosen: (a) resolve and extract

Went with resolve-and-extract over abort-on-include-with-tar_read, since it removes the limitation rather than merely announcing it, and stayed clean to implement:

- `.cdf_extract_include_paths()` — a raw-text regex locating `{{< include ... >}}` shortcodes (both the bare-unquoted and double-quoted forms). This is **not** the `tar_read`-extraction regex the script's header comment rules out — that restriction is about resolving *which targets* a chunk reads (still `purl()`+`parse()`-only, unchanged); this regex only locates *which files* to also `purl()`. Same category as the pre-existing `.cdf_has_r_chunk_fence()`, which is also raw-text regex.
- `.cdf_resolve_include_path()` — resolves relative to the *including file's own directory*, the way Quarto itself resolves includes (not relative to the top-level page or the repo root). Per `.claude/rules/fail-loud-not-null.md`, a dead include target `cli_abort()`s — it is never silently skipped, since a broken include fails at Quarto render time too.
- The old `.cdf_extract_qmd_targets()` per-file purl+parse logic moved unchanged into a new `.cdf_purl_and_extract()`. `.cdf_extract_qmd_targets()` now wraps it: extracts the page's own targets, resolves every include directive, and recurses the same extraction into each included file transitively, unioning the results. A `visited` accumulator (default-valued, invisible to existing callers) guards against a circular include graph recursing forever.

## Demonstrations

**1. It fires.** Created a temporary `docs/_includes/tmp_demo_dead.qmd` with a `tar_read("no_such_target_xyz")` chunk, referenced it from `docs/index.qmd` via a temporary `{{< include >}}` line, and ran Check 1 against the real tree:

```
=== Check 1: dead target references ===
FAIL: dead target references found (these WILL fail at render/tar_read time):
  [DEAD-REF] index.qmd -- references non-existent target(s): no_such_target_xyz
```

Previously this reference was completely invisible — `index.qmd` purls with no fenced chunks of its own containing it, and the include was never resolved. Reverted; confirmed `git status --porcelain` and `git diff --stat` both clean afterward (no leftover temp file, no residual edit).

**2. A live reference in an included file is now counted.** Repeated with a real, currently-unreferenced target (`vix_daily`, picked by diffing `tar_manifest()$name` against the current referenced set) instead of a dead one:

```
=== Check 1: dead target references ===
PASS: no dead target references across 15 page(s), 139 distinct target reference(s)
```

Baseline (unmodified tree, confirmed both before and after the two demos): `PASS: no dead target references across 15 page(s), 138 distinct target reference(s)` — the count is 138 in this PR (up from before #695 landed) and rises to 139 only when a page's include tree gains a real, previously-uncounted reference. Reverted; confirmed clean again.

**3. No false positive.** The unmodified tree — including the real `build-info-footer.qmd` include — still `PASS`es Check 1 at exactly 138, both as a standalone run and inside the full `scripts/verify.sh`.

## Timing

Check 1's `tar_manifest()` cost dominates (~6.7-8.9s across runs in this PR, same order as the pre-existing ~7.5s documented in the script's own header) — extraction itself (purl+parse across all 15 pages, now including the one real include) measured **0.30-0.38s**, materially unchanged from before this PR (there is exactly one real include in the repo today, so the added recursion does one extra `purl()` call).

## Tests

Extended `tests/testthat/test-dashboard-freshness.R` with fixed-basename synthetic fixtures under `withr::local_tempdir()` (`toy_parent.qmd`, `toy_child.qmd`, `toy_grandchild.qmd`) covering:

- `.cdf_extract_include_paths()`: bare-unquoted form, double-quoted form, no-include case, multiple includes.
- `.cdf_resolve_include_path()`: resolves relative to the including file's directory; aborts (snapshot) on a missing target.
- `.cdf_extract_qmd_targets()`: sees a `tar_read()` that lives only in an included file; sees one through a **nested** include chain (parent → child → grandchild); unions the parent's own targets with its include's; aborts (snapshot) on a dead include target; does not infinite-loop on a **circular** include graph (parent ↔ child, each referencing a real target — both targets still collected exactly once).

Every new `cli_abort()` has `expect_snapshot(error = TRUE, ...)` coverage per `.claude/rules/snapshot-test-policy.md`. Caught and fixed one bug via the snapshot itself during development: my first draft of `.cdf_resolve_include_path()`'s abort message interpolated the resolved absolute path (which lives under a volatile tempdir in tests) — non-portable across machines/CI runs. Fixed to show only `raw_path` and `basename(including_path)`, matching this file's existing basename-only convention (see header comment / `portable-build-artifacts`). Also caught a `cli`/glue brace-escaping bug: writing the literal shortcode `{{< include >}}` inside a `{.code ...}` cli interpolation loses a brace pair (glue treats `{{`/`}}` as literal-brace escapes) — reworded to avoid it rather than deal with quadruple-brace escaping.

Also extended the function-signature-stability block (`.cdf_extract_qmd_targets` gained a `visited = character(0)` parameter; added `.cdf_extract_include_paths`, `.cdf_resolve_include_path`, `.cdf_purl_and_extract`).

## Verification

`scripts/verify.sh` (foreground, full mode):

```
::VERIFY:ROOT_SUMMARY:: total=636 failed=3 skipped=0
::VERIFY:PKG_SUMMARY:: total=695 failed=1 skipped=15
```

Root suite total rose from the documented baseline of 625 to 636 (+11, from the new tests above) — failure signatures match the known baseline exactly (`test-crypto-momentum.R::C1`, `test-qa-summary-deps.R::qa_summary declares every *_metrics target...`, `test-stock-backtest.R::raw POSIXct vs Date comparison...`), skip count unchanged at 0. Package suite is untouched by this PR and matches its baseline exactly: total=695, failed=1, skipped=15.

## Scope

This PR is detection-only, per #695's Option A (same as the parent script). It does not touch the re-render/publish half of #695 — that is out of scope here and is being handled separately.

Refs #695

### DIFF
--- a/scripts/check_dashboard_freshness.R
+++ b/scripts/check_dashboard_freshness.R
@@ -79,7 +79,10 @@
 #     ZERO targets -- confirmed legitimate (a landing page with no
 #     tar_read-family calls at all), and is NOT flagged, because the
 #     fence-vs-zero-expression check (not a zero-target-reference check) is
-#     what distinguishes "broken" from "genuinely nothing here".
+#     what distinguishes "broken" from "genuinely nothing here". This
+#     per-file purl()+parse() step now lives in .cdf_purl_and_extract();
+#     .cdf_extract_qmd_targets() wraps it with include resolution (see
+#     "QUARTO {{< include >}} RESOLUTION" below).
 #   - Every tar_read()/tar_read_raw() call's first argument across all 15
 #     files is a character literal or a bare symbol (0 calls with the target
 #     name in a variable/expression) -- confirmed via
@@ -103,6 +106,37 @@
 # non-empty for falsification.qmd specifically ("cg_dag", "cg_caption" --
 # see the inline-expression limitation below), so the wrapper cannot be
 # ignored without a real coverage gap.
+#
+# QUARTO {{< include >}} RESOLUTION (closed 2026-08-20, #695 follow-up) --
+# knitr::purl() tangles only the fenced chunks physically present in the
+# file it is HANDED. It does not resolve Quarto's `{{< include path.qmd >}}`
+# shortcode (that expansion happens inside quarto render's own pandoc
+# preprocessing, never inside knitr), so a tar_read() living in an included
+# file was previously invisible to Check 1 and Check 3 -- SILENTLY: no
+# error, no log line, nothing. Verified empirically 2026-08-20 with a
+# synthetic parent.qmd containing only `{{< include child.qmd >}}` (child.qmd
+# has one fenced `tar_read()` chunk): knitr::purl(parent.qmd) tangles to an
+# EMPTY file (zero expressions), and .cdf_has_r_chunk_fence(parent.qmd) is
+# FALSE (the fence lives in child.qmd, not parent.qmd) -- so the old
+# zero-expressions-but-no-fence branch treated it exactly like index.qmd's
+# legitimate "no code here" case and never flagged the hidden tar_read().
+# .cdf_extract_qmd_targets() now resolves every `{{< include >}}` directive
+# in a page (.cdf_extract_include_paths(), a raw-text regex -- Quarto
+# shortcodes are not knitr chunks, so there is no purl()-based way to find
+# them; this is NOT the tar_read-extraction regex ruled out above, it only
+# locates WHICH FILES to also purl()) and recurses the SAME
+# purl()+parse()-based .cdf_purl_and_extract() into each one, transitively,
+# unioning target references into the including page's set. A `visited`
+# accumulator prevents a circular include graph from recursing forever.
+# Per .claude/rules/fail-loud-not-null.md, an include directive whose target
+# file does not exist is a hard cli_abort() (.cdf_resolve_include_path()),
+# never a silent skip -- a dead include fails at Quarto render time too, so
+# treating it as "nothing to see here" would just move the surprise later.
+# Confirmed against the real repo 2026-08-20: docs/_includes/ contains
+# exactly one file (build-info-footer.qmd), it has zero tar_read calls, and
+# exactly one page (docs/macro-defense-rotation.qmd:710) includes it -- so
+# this gap is provably empty today; the fix exists so the NEXT include that
+# adds a tar_read() does not repeat the silent-miss pattern.
 #
 # KNOWN LIMITATION -- inline R expressions (single-backtick `` `r { ... } ``
 # `` syntax, as opposed to fenced ```{r} chunks) are NOT tangled by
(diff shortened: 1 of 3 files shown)


JohnGavin:
Reviewed and merging. You chose (a) resolve-and-extract, which is the right call —
it removes the limitation rather than announcing it.

I confirmed the traversal is real rather than a no-op, which is the one thing the
138→139 demo alone does not establish:

```
includes found:  _includes/build-info-footer.qmd
resolved:        build-info-footer.qmd
own targets: 14   with-includes: 14
```

Detected, resolved to the right file, recursed into — and correctly contributing
nothing, because that footer has no `tar_read`. A silent skip would also have
printed 14, so this plus your live-reference demo pins it from both sides.

`.cdf_resolve_include_path()` aborting on a missing target is right, and the note
about deliberately showing `raw_path` + `basename()` rather than the resolved
absolute path is the correct reading of `portable-build-artifacts` — a tempdir in
a snapshot would have been a slow-burning CI failure.

One observation, not a defect: `visited` is threaded down each branch but not
shared across siblings, so a diamond (A includes B and C, both including D) would
purl D twice. Results are unioned and `unique()`d, so it is correctness-neutral —
only redundant work, on a structure that does not exist here. Not worth
complicating the accumulator for.

Catching the `cli`/glue brace-escaping problem is a good spot — a literal
`{{< include >}}` inside a `.code` interpolation losing its braces would have
made the abort message quietly wrong in exactly the situation where someone is
relying on it to be precise.

`scripts/verify.sh` on this branch, my own run: PASS, root `636/3/0`
(+11 over `c59026a`'s 625, all passing), package `695/1/15`, both signatures at
baseline.

## For the record on the `gh` invocation

Your environment note is correct and it was my error — every dispatch prompt this
session specified `env -u GH_TOKEN gh ...`, which works for me but is blocked for
agents by the worktree-isolation guard. `GH_TOKEN= gh ...` is the form that works
in a worktree. I have corrected it in my notes so subsequent dispatches say so;
sorry for the detour.

Issue #699, opened 2026-08-20
macro-defense-rotation.qmd unrenderable since the adjusted -> adjusted_close rename — a dead COLUMN reference in a live target, which no check can see


JohnGavin:
`macro-defense-rotation.qmd` has been unrenderable since the `adjusted` → `adjusted_close` rename — a fourth failure surface that none of our checks can see

## The failure

Re-rendering the stale dashboards for #695, eight of nine succeeded. This one
did not:

```
Quitting from macro-defense-rotation.qmd:144-160 [unnamed-chunk-1]
Error in `mutate()`:
ℹ In argument: `normalised = adjusted/first(adjusted) * 100`.
ℹ In group 1: `ticker = "DBC"`.
Caused by error:
! object 'adjusted' not found
```

`docs/macro-defense-rotation.qmd:149`:

```r
prices <- tar_read(bt_prices) |>
  filter(ticker != "BIL") |>
  group_by(ticker) |>
  mutate(normalised = adjusted / first(adjusted) * 100) |>
  ungroup()
```

but the target's actual schema, read from the current store:

```
bt_prices cols: ticker, date, close, adjusted_close     (1373 rows)
```

The column was renamed to `adjusted_close` under #325. This page has been broken
ever since and nobody knew, because nothing re-renders it. Its published HTML is
dated **2026-06-16**, so the live page is a pre-breakage artifact — the site looks
fine while the source cannot build.

Fix is one identifier. The interesting part is why it survived.

## This is a fourth surface, not one of the three we have been tracking

| # | surface | caught by | would it catch this? |
|---|---|---|---|
| 1 | pipeline structure | `scripts/verify.sh` — `parse()` + `tar_validate()` | No — the `.qmd` is not part of the pipeline |
| 2 | target bodies erroring | `scripts/build.sh` → `check_pipeline_errors.R` | No — `bt_prices` builds fine |
| 3 | dead target *reference* | `check_dashboard_freshness.R` Check 1 | **No — and it correctly said PASS** |
| 4 | **dead *column* reference inside a live target** | **nothing** | — |

Check 1 asks "does every target this page reads still exist in the manifest?"
`bt_prices` exists, so it passes, correctly. What changed is the **schema inside**
the target, which Check 1 has no view of by construction.

That is the same shape as everything else this week — a guard scoped to the thing
that broke last time. Check 1 was built to catch a target that disappears. A
target that keeps its name and changes its columns walks straight past it.

## Detection options

[...]

Issue #325, opened 2026-05-28
Normalise adjusted vs adjusted_close column across datasets (follow-up to #316)


JohnGavin:
## Source

Surfaced while documenting #316 (PR #324). The agent confirmed an actual cross-dataset divergence:

| Dataset | Column name | Origin |
|---------|-------------|--------|
| \`equity_daily\` (Yahoo / HF parquet) | \`adjusted\` | yfinance \`adj_close\` renamed at \`scripts/fetch_equity.py:163\` |
| \`alphavantage_daily\` | \`adjusted_close\` | AV API native, no rename |

This is **not** the schema-consistency win the #316 decision implied — the two datasets use different names, so any join across them needs a rename either way.

## Why this matters

- Cross-dataset analysis (e.g., back-fill AV with Yahoo or vice versa) silently breaks until the caller notices the column mismatch.
- The \`hd_alphavantage()\` docstring now points users at \`dplyr::rename(adjusted = adjusted_close)\` to harmonise — but every caller has to remember to do this.
- The choice in #316 (keep \`adjusted_close\` for schema consistency) only holds *within* the AV registry entry; across the package it's the opposite of consistent.

## Acceptance criteria

- [ ] Pick one canonical name (likely \`adjusted_close\` since it's more explicit and matches the AV path).
- [ ] Rename the producing column in the other dataset (probably \`scripts/fetch_equity.py:163\` change \`adj_close\` → \`adjusted_close\` instead of \`adjusted\`).
- [ ] Update any consumers that grep for the old column (likely a handful of \`R/plan_*.R\` files).
- [ ] Update tests + \`inst/COLUMN_NAMING.md\` to reflect the single canonical name.
- [ ] Remove the cross-dataset rename example from \`hd_alphavantage()\` docstring (no longer needed).

## Related

- #316 (closed) — original schema-choice issue
- PR #324 — documented the current divergence
- \`inst/COLUMN_NAMING.md\` — has the canonical table this issue would simplify

PR #701, opened 2026-08-20
fix(macro-defense-rotation): adjusted -> adjusted_close (#325 rename)

JohnGavin:
## Summary

Source fix only — `docs/macro-defense-rotation.qmd:149` referenced a bare
`adjusted` column that has not existed since the #325 rename. No re-render, no
`.html` committed; that happens separately in the main checkout, which has the
`docs/_targets` store this worktree does not.

## How I verified the schema without a store

I cannot read `docs/_targets` from a worktree (no store here), so I verified
from source instead of trusting the issue's claim:

- `R/plan_backtest.R:93-105` — the `bt_prices` target ends with
  `dplyr::select(ticker, date, close, adjusted_close)`. That is the live
  producing code, not a snapshot.
- `R/plan_backtest.R:110-116` — the very next target, `bt_returns`, consumes
  `bt_prices` and correctly references `adjusted_close` (`ret = adjusted_close /
  dplyr::lag(adjusted_close) - 1`), confirming the rename is already in effect
  package-wide and `bt_prices` is not mid-migration.

This matches the issue's reported live schema (`ticker, date, close,
adjusted_close`, 1373 rows) exactly, from two independent source locations.

## The fix

`docs/macro-defense-rotation.qmd:149`:

```r
# before
mutate(normalised = adjusted / first(adjusted) * 100) |>
# after
mutate(normalised = adjusted_close / first(adjusted_close) * 100) |>
```

## Repo-wide search (issue steps 3 and 4)

**Step 3 — other bare `adjusted` column references in `docs/*.qmd`:**
`grep -rnE "\badjusted\b" docs/*.qmd` (word-boundary, so `adjusted_close` does
not self-match) across all 15+ dashboard files. Only two files' worth of hits:
`macro-defense-rotation.qmd:149` (the code reference, now fixed) and
`macro-defense-rotation.qmd:682-683`, plus scattered hits in
`avoid-worst-days.qmd`, `falsification.qmd`, `evidence.qmd`, `leaderboard.qmd`,
`index.qmd`, `jst-dashboard.qmd`, `quiz.qmd`, `momentum-prepeak.qmd`. Every one
of those is prose — "risk-adjusted", "cost-adjusted", "drawdown-adjusted",
"break-adjusted" — not an R column symbol. **Negative result: no other
`docs/*.qmd` file has this defect.**

**Step 4 — other columns #325 renamed:** #325's own PR (#332) file list and
`packages/historicaldata/inst/COLUMN_NAMING.md` both confirm the rename was
`adjusted -> adjusted_close` only — no other column was in scope. I also cross-
checked against the more recent #669/#670 (which fixed 12 `R/plan_*.R`
consumers of the same #325 divergence: `bt_prices`, `bt_returns`, `etf_daily`,
`etf_monthly`, `aw_daily_returns`, `cb_vs_spy`, `gdn_vs_spy`, `spy_returns`,
`multi_asset_returns`, `equity_with_adv`, `olmar_prices`, `rsc_data`, plus
targets in `plan_factormax.R`/`plan_nyt_sentiment.R`/`plan_quiz.R`/
`plan_vix_macro_overlay.R`). `grep -l` for `tar_read()` of any of those target
names across `docs/*.qmd` returns only `macro-defense-rotation.qmd` — no other
dashboard reads any of the 12 targets #670 touched. **Negative result: no
other renamed column is referenced anywhere in `docs/*.qmd`.**

## What I did NOT verify

`scripts/verify.sh` does not render `.qmd` files, so it cannot prove this fix
actually works — it only proves the pipeline still parses/validates and the
test suites are unchanged. **The only real proof is a successful render**,
which needs the `docs/_targets` store this worktree does not have. I could not
run that end-to-end. Recommend re-rendering
`macro-defense-rotation.qmd` in the main checkout (alongside the other #695
work already in progress there) and confirming the price-history chunk builds
without `object 'adjusted' not found`.

## Verify

`scripts/verify.sh`, foreground, full mode — PASS, baselines unchanged:

```
SKIP count this run: 0
PASS: failures match the known baseline exactly

SKIP count this run: 15
PASS: failures match the known baseline exactly

::VERIFY:ROOT_SUMMARY:: total=625 failed=3 skipped=0
::VERIFY:PKG_SUMMARY:: total=695 failed=1 skipped=15
```

Matches the expected baselines exactly (root `total=625 failed=3 skipped=0`,
package `total=695 failed=1 skipped=15`) — this `.qmd`-only edit moved
neither.

## Out of scope (deliberately not built)

#699 proposes schema snapshots (`expect_snapshot_value()` on dashboard-read
target `names()`) as the durable fix for this class of failure. Not started
here — separate, larger piece of work. One observation for that follow-up: it
should probably snapshot the schema of every target any `docs/*.qmd` actually
`tar_read()`s (the same 138-target surface `check_dashboard_freshness.R`
Check 1 already extracts), not just the 12 targets #670 happened to fix —
otherwise the next drift in an untouched target repeats this exact failure
mode on a different column.

## Test plan

- [x] `scripts/verify.sh` (foreground, full) passes, baselines unchanged
- [x] Repo-wide word-boundary search confirms no other `docs/*.qmd` file has
      a bare `adjusted` column reference
- [x] Cross-checked against every target #670 renamed — only this dashboard
      reads any of them
- [ ] Render `macro-defense-rotation.qmd` in the main checkout (needs the real
      `docs/_targets` store — not available in this worktree)

Refs #699

### DIFF
--- a/docs/macro-defense-rotation.qmd
+++ b/docs/macro-defense-rotation.qmd
@@ -146,7 +146,7 @@ Normalised price history of the 4 defensive ETFs + SPY benchmark since 2007, reb
 prices <- tar_read(bt_prices) |>
   filter(ticker != "BIL") |>  # exclude cash proxy from price chart
   group_by(ticker) |>
-  mutate(normalised = adjusted / first(adjusted) * 100) |>
+  mutate(normalised = adjusted_close / first(adjusted_close) * 100) |>
   ungroup()
 
 n_tickers <- length(unique(prices$ticker))


JohnGavin:
Verified both claims independently before merging.

**Schema, from source:** `R/plan_backtest.R:103` ends
`select(ticker, date, close, adjusted_close)`, and `bt_returns` at :113 already
consumes `adjusted_close`. So this page was the sole straggler from #325 — the
pipeline itself had moved on.

**Negative result, re-run my own way:** a word-boundary search across every
`docs/*.qmd`, excluding the prose compounds, leaves
`macro-defense-rotation.qmd:149` as the only code-level occurrence. Everything
else is "drawdown-adjusted", "rf-adjusted", "adjusted price" in narrative text.

Your cross-check against the full #670/#325 target list is the part that makes
this a fix to the *class* rather than to the line — and reporting the negative
explicitly is what makes it checkable. That was the main thing I wanted from this
dispatch.

## On verification

I am deliberately **not** re-running `scripts/verify.sh` myself here. It does not
render `.qmd` files, so it cannot exercise this change at all; your run confirming
both baselines is enough to show nothing else broke, and a second identical run
would add no information. The real check is a successful render, which I am about
to run in the main checkout. If that fails, this comes back out.

Stating that explicitly rather than performing the ritual, because "I ran the
standard gate" would imply more coverage than this change actually received.

## Environment finding worth keeping

Your `git push` result generalises the one in the dispatch: the ambient invalid
`GH_TOKEN` is read by gh's git-credential helper as well as by the CLI, so
`GH_TOKEN= git push` is needed too — not just `GH_TOKEN= gh`. Added to my notes
alongside the CLI form. The wrong prescription was mine.

Issue #740, opened 2026-08-24
The render gate only renders already-stale pages, and its error-pattern grep false-positives on vendored JS, base64 and case-folded NaN


JohnGavin:
Found while auditing every dashboard to establish what the published site actually
shows. Two independent weaknesses, one that makes the gate blind and one that makes it
cry wolf. Either alone is tolerable; together they mean the gate cannot be trusted in
either direction.

## The good news first

I rendered **all 15** `docs/*.qmd` against the live store. **Every one rendered
successfully, and none has a genuine reader-visible error.** #699 and #530 are both
resolved. The dashboard suite is healthier than the issue list implies.

Establishing that took three passes, because the first two both lied.

## Weakness 1 — the gate only renders pages that are already stale

`scripts/build.sh --render` (#695/#708) parses Step 3's log for `[STALE]` /
`[DATA-STALE]` markers and renders **only those pages**. Today that was 4 of 15:
`evidence`, `falsification`, `leaderboard`, `momentum-prepeak`.

The other 11 are never rendered by any gate. So a page that is **permanently
up-to-date but broken** is never exercised — which is *precisely* what #699 and #530
described. The gate cannot see the class of bug it exists to catch, because its
trigger is staleness rather than correctness.

Check 1 (dead target references) does scan all 15, but statically — it cannot see a
dead *column* reference inside a live target, which was #699's entire point.

**Proposed:** a `--render-all` mode, plus one scheduled full-render (weekly is
probably enough) so every page is exercised on a known cadence regardless of
staleness. The all-pages code path already exists — it is currently reachable only as
a fallback when the freshness marker is missing.

## Weakness 2 — the error-pattern grep false-positives three ways

The verification rule's prescribed check greps rendered HTML for
`not available|not found in targets|MISSING EVIDENCE|Error in |Error:|#&gt;|NaN`.
Applied naively that produced, on today's clean site:

| source of false positive | example | pages affected |
|---|---|---|
| case-insensitive `NaN` | "fi**nan**cial intermediary" | european-overlay, others |
| vendored minified JS | DataTables' `iDrawError:-1` matches `Error:` | macro-defense-rotation (3) |
| base64 data URIs | embedded font blobs containing the literal `NaN` | macro-defense-rotation (5) |

First pass reported 38 hits on `macro-defense-rotation`. Case-sensitive matching cut
it to 18. Stripping `<script>`/`<style>` cut it to 5. Inspecting those 5 showed all
[...]

PR #743, opened 2026-08-24
fix(build): render-gate blind spot + false-positive error grep (#740); add missing build-info footers (#697)

JohnGavin:
## Summary

Two independent fixes, dispatched together because #740 named both:

**Refs #740 — render gate blind spot + false-positive error grep**
- `scripts/build.sh`: new `--render-all` flag forces Step 4 to render every
  `docs/*.qmd` page regardless of Step 3's staleness verdict (reuses the
  existing all-pages fallback path). Writes `docs/_full_render_marker.txt`
  on completion for a scheduled cadence check to read.
- `scripts/check_html_errors.sh` (new): case-sensitive, `<script>`/`<style>`/
  `data:` URI-stripped error-pattern matcher. Reduces the naive 38-hit
  false-positive count on `docs/macro-defense-rotation.html` to the true
  count of zero. `scripts/build.sh` Step 4 now delegates to it.
- `tests/test_check_html_errors.sh` (new): regression fixtures for each
  false-positive source plus a true-positive fixture.
- `.github/workflows/full-render-cadence.yml` (new): weekly job that reads
  the committed marker's age and opens an issue if `--render-all` hasn't run
  recently. Never runs `tar_make()`/`quarto render` itself in CI — same
  reasoning `dashboard-freshness.yml`'s Check 3 already documents.

**Refs #697 — missing build-info footers**
- `docs/avoid-worst-days.qmd`, `docs/bdbb-sol.qmd`: added the build-info
  footer via `{{< include _includes/build-info-footer.qmd >}}` (same include
  `macro-defense-rotation.qmd` uses). Posted a correction comment on #697:
  of the 6 pages originally flagged, 4 are redirect stubs
  `check_dashboard_freshness.R` already excludes — only these 2 needed one.

## What I verified (OBSERVED, not predicted)

- `scripts/verify.sh` (full run, foreground): PASS — 3/3 known root-suite
  baseline failures, 1/1 known package-suite baseline failure, 15 skips
  (matches `.claude/CLAUDE.md`'s documented baselines exactly).
- `scripts/check_html_errors.sh` run against all **15** real rendered
  `docs/*.html` files in the main checkout (read-only): **0 hits on every
  page** — matches #740's own claim that the live site has zero genuine
  errors.
- `scripts/check_html_errors.sh --all` (run from the worktree): correctly
  enumerates the same 15 pages and excludes `MERMAID_LESSONS.html` (no
  corresponding `.qmd` — it renders from `MERMAID_LESSONS.md` and isn't one
  of the 15 dashboard pages this gate covers).
- Reproduced the false-positive counts by hand against the real
  `macro-defense-rotation.html`: naive case-insensitive `NaN` grep = 35 hits,
  naive `Error:` grep (unstripped) = 3 hits — both go to 0 after stripping.
- `bash -n scripts/build.sh`: syntax OK.
- `scripts/build.sh --render-all`: flag is accepted (falls through to the
  pre-existing "refuse to build from a linked worktree" guard, never
  attempted to touch the real store — required, since I'm running from a
  worktree and must not build against the real `docs/_targets`).
- `scripts/build.sh --bogus-flag`: still correctly rejected as unknown.
- `find docs -maxdepth 1 -name '*.qmd' | ...`: the literal selection command
  `--render-all` now forces returns all 15 `.qmd` basenames.
- `tests/test_check_html_errors.sh`: 9/9 assertions pass (3 false-positive
  sources individually suppressed, 1 combined fixture, 1 true-positive
  regression guard, usage errors, `--all` scope).
- `tests/test_build_lock.sh` (pre-existing): 13/13 assertions still pass —
  no regression from the `build.sh` edits.

## What I stubbed / could not do

- **Did not run `tar_make()`** against the real store — forbidden by the
  dispatch and refused by `scripts/build.sh` itself when run from a
  worktree.
- **Did not perform an actual `quarto render`** of any page — that requires
  the real `docs/_targets` store, which does not exist in this worktree, and
  `scripts/build.sh` explicitly refuses to build from a linked worktree.
  Task instructions said this is acceptable ("You do NOT need to actually
  render all 15 (slow)... say what you stubbed"). As the closest available
  substitute I verified `check_html_errors.sh` (the post-render half of Step
  4) against all 15 already-rendered real pages, read-only, in the main
  checkout.
- The weekly full-render-cadence workflow is a **cadence reminder, not an
  actual re-render in CI** — deliberately, for the same reason
  `dashboard-freshness.yml`'s own header comment already gives for its Check
  3: a GH Actions runner has no live data-fetch credentials and no store,
  and building one would race the main checkout's canonical store. This is a
  design trade-off, not an oversight — flagging it explicitly since the
  issue's wording ("weekly GitHub workflow is fine") could be read as asking
  for an actual re-render.
- No `docs/_full_render_marker.txt` exists yet in this PR (nobody has run
  `--render-all` against the real store) — the cadence workflow will
  correctly fail-with-a-clear-message on its first scheduled run until a
  human does so once, which is the intended bootstrap behaviour.

## Design trade-off (favour false positive over false negative)

`check_html_errors.sh` strips `<script>`, `<style>`, and `data:` URI content
before matching — but never touches rendered prose, table cells, or
captions a reader would see, so a genuine defect there is still caught (see
`tests/test_check_html_errors.sh` Test 4/5). `NULL` and bare `NA` were
deliberately **not** added to the pattern list, unlike the fuller table in
`verification-before-completion.md`'s Post-Deploy Validation section — both
are short substrings of very common ordinary words/identifiers ("NAME",
"NATIONAL", "NULLify") with no reliable word/tag boundary, and adding them
unbounded would reintroduce exactly the false-positive class this script
exists to remove.

## Test plan

- [x] `scripts/verify.sh` full run
- [x] `check_html_errors.sh` against all 15 real rendered dashboards
- [x] `check_html_errors.sh --all` selection/exclusion behaviour
- [x] `tests/test_check_html_errors.sh` (new, 9 assertions)
- [x] `tests/test_build_lock.sh` (pre-existing, no regression)
- [x] `bash -n scripts/build.sh`
- [x] `--render-all` / unknown-flag argument parsing

Refs #740, #697.

### DIFF
--- a/.github/workflows/full-render-cadence.yml
+++ b/.github/workflows/full-render-cadence.yml
@@ -1,0 +1,111 @@
+name: Full render cadence (weekly)
+
+# #740: scripts/build.sh --render (no flag) only ever renders pages Step 3
+# flags [STALE]/[DATA-STALE] -- a page that is PERMANENTLY up-to-date but
+# broken is therefore never exercised by any gate. `scripts/build.sh
+# --render-all` closes that gap by rendering every docs/*.qmd page
+# unconditionally, but it is a manual, ~20-minute human act (same cost
+# dashboard-freshness.yml's header comment already documents for an ordinary
+# --render) -- nothing enforces that anyone actually runs it periodically.
+#
+# WHY THIS WORKFLOW NEVER RUNS quarto render ITSELF: doing the real thing in
+# CI would mean running tar_make() against docs/_targets.R on a fresh
+# runner -- live external data fetches (equity/macro/crypto APIs, credentials
+# this workflow does not have), a 10+ minute cold nix build, and a store that
+# is NOT the main checkout's (see .claude/CLAUDE.md and the
+# worktree-location rule for why a second, unofficial store must never race
+# the canonical one -- scripts/build.sh itself refuses to run anywhere except
+# the main checkout for exactly this reason). dashboard-freshness.yml's own
+# header comment already rejected this same approach for its Check 3 and
+# instead reads a small COMMITTED snapshot; this workflow follows the same
+# pattern for the same reason.
+#
+# So this workflow reads ONE committed file, docs/_full_render_marker.txt --
+# written by `scripts/build.sh --render-all` (never committed automatically;
+# a human reviews and commits it exactly like any other file that script
+# writes under docs/) -- and fails if it is older than
+# FULL_RENDER_MAX_AGE_DAYS, or missing entirely. A green run here means
+# "someone ran --render-all recently enough that every dashboard page has
+# been exercised, not just the ones Step 3 happened to flag as stale."
+#
+# WEEKLY, NOT DAILY: mirrors dashboard-freshness.yml's own cadence
+# reasoning -- the fix is a deliberate ~20-minute human act, and a threshold
+# measured in weeks gives a full working week to act before this goes red.
+
+on:
+  schedule:
+    - cron: '30 8 * * 1'      # weekly, Monday 08:30 UTC (after dashboard-freshness.yml's 08:00 run)
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  issues: write               # to auto-file an issue on failure
+
+env:
+  # How old docs/_full_render_marker.txt may be before this workflow fails.
+  # 30 days: roughly monthly, consistent with the ~20-minute human cost of a
+  # full render and this repo's other staleness thresholds being measured in
+  # weeks, not days (see dashboard-freshness.yml's own cadence comment).
+  FULL_RENDER_MAX_AGE_DAYS: 30
+
+jobs:
+  full-render-cadence:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Check docs/_full_render_marker.txt age
+        id: check
+        run: |
+          REPORT="./full-render-cadence-report.md"
+          MARKER="docs/_full_render_marker.txt"
+          if [[ ! -f "$MARKER" ]]; then
+            {
+              echo "\`$MARKER\` not found -- no \`scripts/build.sh --render-all\` has ever"
+              echo "been committed."
+              echo ""
+              echo "Run \`scripts/build.sh --render-all\` in the main checkout, review the"
+              echo "output, and commit the resulting \`$MARKER\` (and any refreshed"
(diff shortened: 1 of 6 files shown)


JohnGavin:
Superseded by #744, now merged. #744 carries this PR's commit (`c41d96f`) plus a fix
for a regression this branch introduced.

## What I found by running the thing this PR adds

`scripts/build.sh --render-all` against the real store: **14 pages rendered clean,
`bdbb-sol.qmd` failed**, exit 3.

```
12/13 [build-info]
Error in `build_info()`: could not find function "build_info"
Quitting from bdbb-sol.qmd:103-108 [build-info]
```

The #697 half added `{{< include _includes/build-info-footer.qmd >}}` to two pages. The
include is a bare chunk calling `build_info()`, which is defined in
`docs/vignette_utils.R`. `avoid-worst-days.qmd` happens to `source()` that file;
`bdbb-sol.qmd` never did. So the include carried an **undeclared prerequisite** — it
worked only on pages that had already sourced something it never mentioned.

Worth stating plainly: **`--render-all` caught a bug introduced by the same PR that
added it.** `bdbb-sol` was not stale, so the pre-#740 gate would never have rendered it
and this would have shipped. That is a better argument for this PR than anything in its
description.

## Why the fix went into the include, not the page

Adding `source()` to `bdbb-sol.qmd` would have fixed the instance and left the trap
armed — and #697 exists precisely *because* pages are still missing this footer, so
more pages will get the include. #744 makes the include self-sufficient
(`exists("build_info", mode = "function")` guard, then the established
local-path-then-`here::here()` idiom). Fixing the property rather than the instance,
per `.claude/rules/fail-loud-not-null.md` Required Pattern 5.

## The rest of this PR verified well

I tested the new matcher in **both** directions rather than trusting a zero — a matcher
tuned until a clean site reports nothing could simply have gone blind:

| test | result |
|---|---|
| all 15 real rendered pages | 0 hits, exit 0 — matches my own independent audit |
| a copy with 3 genuine errors injected into visible DOM | caught all 3, exit 1 |

After #744, end-to-end on the real store: **777 targets / 0 errored, 15/15 pages
rendered clean, exit 0**, and the build-info footer confirmed present on `bdbb-sol`,
`avoid-worst-days`, `leaderboard` and `macro-defense-rotation` — and correctly absent
from `drif`, which is a redirect stub.

Closing in favour of #744. Nothing here is lost; it is all in `main`.

Refs #740, #697, #744.

Issue #738, opened 2026-08-23
CMR's series is monthly before 2000 and daily after, declared daily throughout — and the median-gap guard cannot see it


JohnGavin:
Surfaced by #736's resampling diagnostic, which reported *"cmr: excluded 91 partial
months from the monthly-resampled spine ... fewer than 15 trading days"* and named
every one. Following that up produced a hard measurement.

## The measurement

`cmr_portfolio_1m`, observations per year and median gap between consecutive dates:

| era | obs/year | median gap |
|---|---:|---:|
| 1992–1999 | **12** | **31 days** |
| 2000–2026 | ~255 | **1 day** |

n = 6852 total. **94 observations are monthly-spaced; 6758 are daily.** The series
changes frequency in January 2000 and is declared `daily` / `ann_factor = 252` for its
entire length (`STRATEGY_OBS_ANN_FACTOR`, `R/plan_strategy_names.R` position 11).

This is the "CMR mixes daily and monthly observations" concern that was deliberately
fenced off from #724/#725 as needing its own decision. It now has evidence and a
magnitude.

## Why the existing guard does not catch it

`.assert_cmr_ann_factor()` (added by #720, the fix for #717) reconciles a declared
`ann_factor` against the observed frequency using the **median gap** between
observation dates, in bands (≤3 → 252, ≤10 → 52, ≤45 → 12, ≤135 → 4).

The overall median gap for CMR is **1 day**. 94 monthly observations out of 6852
cannot move a median. **The guard passes, correctly by its own logic, on a series
where 1.4% of the observations are at a twenty-one-fold different frequency.**

This is the recurring shape again, one level up from where it was last fixed: the
guard is scoped to a **summary statistic** rather than to the property. A median gap
answers *"what is the typical spacing"*. The property we actually want is *"is the
spacing consistent"* — and a median is precisely the statistic chosen to be
insensitive to the minority that would reveal the answer is no.

Compare #710, #694, #691: guard scoped around the known instance rather than the
property. Here it is scoped around the central tendency rather than the dispersion.

## Likely materiality

Not yet measured, but the arithmetic says it is not negligible. A monthly return has
roughly √21 the standard deviation of a daily one, so ~21× the variance. Pooling:

- 94 monthly obs × ~21 units of variance ≈ 1974
- 6758 daily obs × 1 unit ≈ 6758

So on order of **23% of CMR's pooled return variance comes from 1.4% of its
observations**, all of which are then annualised as though they were daily. That
[...]

Issue #717, opened 2026-08-22
CMR is daily data annualised as monthly — vol understated ~4.6x, CAGR annualised over 566 years, and it fed the sigma=11.5% leverage recommendation


JohnGavin:
## The defect

`.compute_cmr_metrics()` (`R/plan_commodities_mean_reversion.R:204`) takes
`ann_factor = 12L` and every call site passes `12L` (lines 98, 102, 106). The file
header asserts it too: `# Frequency: monthly (ann_factor = 12).`

**The data is daily.** Measured against `cmr_portfolio_1m` in the live store:

```
rows: 6852
range: 1992-03-01 to 2026-05-14   (span 34.2 years)
median gap between dates: 1 day
observations per year: 200.3
```

6852 observations over 34.2 years is ~200/year — business-daily with gaps. It is
not 6852 months, which would be 571 years.

## What that breaks

Inside `.compute_cmr_metrics()`, with `n = 6789` and `ann_factor = 12`:

| line | computation | effect |
|---|---|---|
| `years <- n / ann_factor` | 6789 / 12 = **565.75 years** | CAGR annualised over 566 years instead of ~27 |
| `vol <- sd_r * sqrt(ann_factor)` | `sqrt(12)` where `sqrt(252)` is needed | **vol understated by ~4.58×** |
| `sharpe_ratio_rf(..., periods_per_year = ann_factor)` | 12 | Sharpe wrong on both numerator and denominator |

Published CMR Full Period figures are therefore all wrong:
`cagr = -0.0078`, `vol = 0.0499`, `sharpe = -0.544`.

A commodities mean-reversion strategy running at **5.0% annualised vol** should have
looked implausible on its own. It is the second-lowest vol on the entire
leaderboard.

This also explains the `n_months` naming: `n_months = n` at line 253 labels a daily
count as months, which is how it reaches the leaderboard's `months` column via
`.norm_cmr`'s `months = n_months`. Three other strategies (OLMAR-1, TOM, Avoid
Worst) also carry daily counts in `months`, but they do so through an explicit
`rename(months = n_days)` and their *annualisation* is correct — only their column
label is misleading. CMR is the one where the wrong periodicity reaches the
arithmetic.

## It propagated into the leverage recommendation

This is the part that matters beyond the leaderboard. CMR's understated vol gave it
the second-lowest `vol_per_unit_gross` (0.0250) in [#635](https://github.com/JohnGavin/historical/issues/635), and hence
[...]

PR #748, opened 2026-08-24
fix(cmr): periodicity consistency check + measured materiality of CMR's mixed-frequency series (#738)

JohnGavin:
Two halves, per the dispatch. **Half 1 (measurement) is the deliverable**; Half 2 fixes the guard. **No remedy is applied to CMR's data** — that decision stays open on #738 and should follow these numbers.

Everything below is OBSERVED from the live store (`docs/_targets`, read-only via `tar_read()`). Nothing is predicted, and nothing is scaled by hand. `tar_make()` was not run.

---

# Half 1 — Materiality, measured

## Headline: the issue's ~23% estimate is wrong by a factor of ~4.7. The real figure is 4.9%.

The issue reasoned that a monthly return carries ~21× the variance of a daily one, giving `94 × 21 / (94 × 21 + 6758) ≈ 23%`. That step assumes the pre-2000 monthly observations are a correctly-scaled monthly aggregation of the post-2000 daily process. **They are not.**

| lookback | monthly-spaced obs | % of obs | **% of pooled SS** | sd (monthly obs) | sd (daily obs) | ratio | ratio if √21 |
|---|---:|---:|---:|---:|---:|---:|---:|
| 1m | 95 | 1.386% | **5.92%** | 0.02745 | 0.01294 | 2.12 | 4.58 |
| 3m | 93 | 1.358% | **5.47%** | 0.02807 | 0.01392 | 2.02 | 4.58 |
| 6m | 91 | 1.330% | **4.88%** | 0.02750 | 0.01429 | 1.92 | 4.58 |

The monthly observations carry only ~2× the sd of the daily ones, not ~4.6×.

**Why — and this is a separate finding worth its own issue.** Before 2000 the commodity universe is 13 FRED monthly indexes and nothing else; Yahoo's 24 daily series start in 2000:

| era | source | rows | series | series available per date (min / med / max) |
|---|---|---:|---:|---|
| pre-2000 | fred_imf | 1,235 | 13 | 13 / 13 / 13 |
| 2000+ | fred_imf + yahoo | 144,889 | 37 | 1 / 23 / 37 |

`hd_commodity_mr_portfolio()` is called with `n_long = 10, n_short = 10` — **20 slots drawn from 13 names**. Pre-2000 the long and short legs necessarily overlap heavily, so the legs largely cancel and the portfolio return is mechanically muted. The pre-2000 segment is not a monthly version of the strategy; it is a *different, largely self-cancelling* strategy that happens to share a name.

## The published figures (6m — the lookback `.norm_cmr()` selects, matching the deployed leaderboard row)

| metric | published (whole mixed series, ann 252) | 2000+ daily only, ann 252 | 1992–1999 monthly only, ann 12 | published vs coherent |
|---|---:|---:|---:|---|
| `vol` | **0.2288** | 0.2245 | 0.0958 | **overstated 1.92%** |
| `cagr` | **−0.1515** | −0.1669 | +0.0597 | **flattered 9.23%** (less negative) |
| `sharpe` | **−0.747** | −0.829 | +0.602 | **flattered 9.89%** (less negative) |
| `max_dd` | −0.9956 | −0.9956 | −0.1653 | unchanged |
| n | 6,789 | 6,700 | 89 | |

All three lookbacks, for completeness:

| lookback | vol mixed → 2000+ | Δ% | cagr mixed → 2000+ | Δ% | sharpe mixed → 2000+ | Δ% |
|---|---|---:|---|---:|---|---:|
| 1m | 0.2085 → 0.2035 | +2.46 | −0.3688 → −0.3724 | −0.97 | −1.862 → −1.924 | −3.22 |
| 3m | 0.2246 → 0.2198 | +2.18 | −0.2567 → −0.2705 | −5.10 | −1.230 → −1.317 | −6.61 |
| 6m | 0.2288 → 0.2245 | +1.92 | −0.1515 → −0.1669 | −9.23 | −0.747 → −0.829 | −9.89 |

### Plainly: overstated, understated, or unchanged?

- **`vol` is overstated, but immaterially** — 1.9% relative, 0.0043 absolute. The direction is as the issue guessed; the magnitude is not. This is nothing like #717's 4.6×.
- **`cagr` and `sharpe` are flattered by ~10%** — CMR looks meaningfully *less bad* than a coherent series says it is. That is the opposite sign to `vol` and it is the larger effect. It is not an annualisation artefact: the pre-2000 segment genuinely earned +6.0% CAGR at Sharpe +0.60, because (see above) it was a near-self-cancelling 13-name portfolio, and those 7.75 years are averaged into the full-period figure.
- **`max_dd` is unchanged** at −0.9956. CMR destroys essentially all capital in both the mixed and the coherent series.

**The direction does not cancel out, and it is not a single direction.** The mixing has two channels pulling opposite ways, which is exactly why this needed measuring.

### The largest arithmetic error is one the issue did not name

`years <- n / ann_factor` inside `.compute_cmr_metrics()` counts each of the 91 monthly observations as *one day*:

| lookback | n | years assumed (n/252) | actual calendar span | understated by | cagr at assumed years | cagr at true span |
|---|---:|---:|---:|---|---:|---:|
| 1m | 6,799 | 26.98 | 33.99 | **7.01 yr (26.0%)** | −0.3688 | −0.3060 |
| 3m | 6,795 | 26.96 | 33.83 | **6.86 yr (25.4%)** | −0.2567 | −0.2106 |
| 6m | 6,789 | 26.94 | 33.57 | **6.63 yr (24.6%)** | −0.1515 | −0.1235 |

Every published CMR CAGR is annualised over ~27 years for a series spanning ~34. This is a ~25% error in the exponent and it is bigger than either effect in the table above. It pulls CAGR *more* negative, partly offsetting the flattering from the benign pre-2000 era.

## The #635 channel — what a wrong vol does to a position size

`vol_per_unit_gross = vol / gross_convention`; `G = σ_target / vol_per_unit_gross`. CMR's `gross_convention` is **2** (`is_cap = FALSE`).

| | vol | `vol_per_unit_gross` | rank by vpug |
|---|---:|---:|---|
| published (mixed) | 0.2288 | 0.11440 | 12th of 15 |
| coherent (2000+) | 0.2245 | 0.11225 | 12th of 15 |

`G_cmr` would be **1.92% larger** under the coherent vol — the overstated vol makes CMR's implied gross allocation slightly too *small*. Book-level, budget-neutral σ (`Σ Gᵢ = Σ grossᵢ`, Full Period):

| scenario | σ |
|---|---:|
| leaderboard as deployed (CMR vol 0.2288) | 0.13541 |
| CMR vol → 2000+ coherent (0.2245) | 0.13529 |
| CMR excluded entirely | 0.13077 |

**σ moves by −0.09%.** For scale: #717's defect moved the same quantity by ~17%. This one is three orders of magnitude smaller and is inside any sensible rounding. **The mixed-frequency series is not a threat to the #635 leverage number.** It *is* a threat to CMR's own CAGR and Sharpe, and to anyone reading the pre-2000 era as evidence about the strategy.

## Recommendation (reasoning, not a change)

**Option 1, truncate to 2000+** — with the caveat that the case for it is not the periodicity at all.

The periodicity argument alone would not justify losing 7.5 years for a 1.9% vol correction. What justifies it is the universe: pre-2000 the strategy cannot be run as specified (20 slots, 13 names). Those 94 observations are not weak evidence about CMR — they are evidence about a *different* portfolio, and they currently flatter CMR's published CAGR and Sharpe by ~10% while shrinking its annualisation span by 25%.

Against Option 2 (resample everything to monthly): it would change the strategy's construction, not just its reporting, and discards the daily resolution the post-2000 signal is built on — a large cost to fix a 1.9% vol error. Against Option 3 (segment and report separately): most honest, but a leaderboard row has one `vol` cell, and given that CMR's max drawdown is −99.6% in both eras, the reporting sophistication is unlikely to earn its keep.

Worth deciding separately and probably worth its own issue: **the `n_long = 10, n_short = 10` parameters are not satisfiable on a 13-name universe.** Truncating to 2000+ hides that rather than fixing it, and the same trap will reappear on any date range where the universe is thin.

---

# Half 2 — The guard

## What changed

`.assert_cmr_ann_factor()` keeps the #720 median-gap **classification** check (that check answers "what is the typical spacing" and answers it correctly) and gains a second, **consistency** check that answers "is the spacing consistent with *one* declared periodicity". That is a question about dispersion, and no measure of central tendency can answer it.

Concretely: count the gaps falling outside the declared periodicity's calendar band; abort when the count exceeds an allowance.

### The bands (`CMR_PERIODICITY_TOLERANCE`) — every bound argued in its roxygen

| declared | nominal spacing | band | why |
|---|---:|---|---|
| 252 daily | 1.45 d | **[1, 10]** | Longest US market closure since 1990 is 9/11 — last print 2001-09-10, next 2001-09-17, a **7-calendar-day gap**. Christmas/New-Year adjacent to a weekend reaches 4–5; Sandy (2012) gave 4. 10 clears the worst observed with headroom and sits three weeks below a monthly spacing, so a monthly observation can never pass as a long holiday. Lower bound 1: a date-keyed series cannot be denser than one obs/day. |
| 52 weekly | 7.02 d | [4, 24] | two consecutive missing weeks tolerated; sub-half-week spacing rejected |
| 12 monthly | 30.44 d | [20, 75] | 20 clears a 28-day February; 75 tolerates one entirely missing month (2×31) and stays below quarterly |
| 4 quarterly | 91.31 d | [60, 200] | one missing quarter tolerated; monthly spacing rejected |

The check is **two-sided**: a gap too *short* for the declared periodicity is the same defect as one too long (daily prints leaking into a monthly series).

Adjacent bands overlap by design — a 7-day gap is consistent both with a daily series over a holiday week and with a weekly series. That ambiguity affects only the diagnostic label in the message; the check itself only ever asks whether a gap lies in the *declared* band, so overlap cannot cause a miss or a false positive. Pinned by a test.

### The allowance — not a magic number

```
allowance = max(2, ceiling(0.001 * n_gaps))
```

**Why 0.1%:** the smallest frequency *regime change* worth catching is one year of a different periodicity embedded in a longer series — 12 monthly gaps inside a ~27-year daily series (~6,800 gaps) is 0.18%, roughly 2× the threshold. Anything below 0.1% is an isolated vendor outage, not a change of periodicity.

**Why not zero:** a naive "all gaps identical" or "zero out-of-band" rule false-positives on every real series. This is not hypothetical — the sweep below found `daily_rf` carries **one genuine 12-day gap in 26,189**. A guard that aborts the whole pipeline on one bad print gets switched off rather than fixed.

**Why the floor of 2:** on a short series `ceiling(0.001 × n)` rounds to 1, so a single outage would abort. The floor keeps the guard firing on a *pattern*, never on one print, at any series length.

### Abort message

Names the strategy, the declared factor, the observed band counts, the out-of-band gap range and its date span, and why the median-gap check passed — per `fail-loud-not-null.md`:

```
x CMR 1m: the observation spacing is NOT consistent with a single declared periodicity (ann_factor 252, daily).
i 94 of 6851 gaps (1.372%) fall outside the daily band [1, 10] calendar days; allowance is 7.
i Observed gap bands: daily: 6757; monthly: 94.
i Out-of-band gaps range 28-31 calendar days, spanning 1992-04-01..2000-01-01.
i The median gap (1 calendar days) agrees with the declared factor, which is why the
  #720 median-gap check passes -- a median cannot see a minority at a different frequency. See #738.
i See .claude/rules/fail-loud-not-null.md Required Pattern 5.
```

## It fires on CMR as it stands today — that is the point

Run against the live store (read-only):

| series | declared | n | median gap | out-of-band | allowance | result |
|---|---:|---:|---:|---:|---:|---|
| `cmr_portfolio_1m` | 252 | 6,852 | 1 | 94 | 7 | **ABORTS** |
| `cmr_portfolio_3m` | 252 | 6,848 | 1 | 92 | 7 | **ABORTS** |
| `cmr_portfolio_6m` | 252 | 6,842 | 1 | 90 | 7 | **ABORTS** |
| 2000+ segment | 252 | 6,758 | 1 | 0 | 7 | passes |
| pre-2000 segment | 12 | 94 | 31 | 0 | 2 | passes |

Both coherent segments pass, so the guard is detecting the *mixing*, not merely rejecting CMR.

## Deliberate cross-strategy sweep — CMR is alone

#738 notes that #736's evidence was a by-product of one diagnostic. This is the deliberate version, applying the identical band + allowance rule to every daily-declared series in the store:

| series | n | median gap | max gap | out-of-band | allowance | result |
|---|---:|---:|---:|---:|---:|---|
| `cmr_portfolio_1m` | 6,852 | 1 | 31 | 94 | 7 | **would fire** |
| `cmr_portfolio_3m` | 6,848 | 1 | 31 | 92 | 7 | **would fire** |
| `cmr_portfolio_6m` | 6,842 | 1 | 31 | 90 | 7 | **would fire** |
| `olmar_portfolio` | 4,063 | 1 | 5 | 0 | 5 | ok |
| `tom_portfolio` | 8,092 | 1 | 7 | 0 | 9 | ok |
| `tom_daily` | 8,092 | 1 | 7 | 0 | 9 | ok |
| `rsc_portfolio` | 8,355 | 1 | 7 | 0 | 9 | ok |
| `aw_daily_returns` | 8,355 | 1 | 7 | 0 | 9 | ok |
| `daily_rf` | 26,190 | 1 | 12 | 1 | 27 | ok |
| `aw_daily_ff` | 26,190 | 1 | 12 | 1 | 27 | ok |

Note the `max gap = 7` on four independent series — that is the 9/11 closure, the exact calendar fact the 10-day daily bound was set from. And `daily_rf`'s single 12-day gap is the empirical case the allowance floor exists for.

## ⚠️ This WILL break the build. It has not been softened.

`.assert_cmr_ann_factor()` runs at the top of `.compute_cmr_metrics()`, so on the next `tar_make()` it aborts **`cmr_metrics_1m`, `cmr_metrics_3m`, `cmr_metrics_6m`** and everything downstream — `cmr_summary`, `cmr_vs_mom_compare`, `cmr_registry_run`, the leaderboard, and the QA gates that read it.

That is the correct behaviour and the proof the guard works. I have not weakened it.

**Proposed landing path — your call, I have not taken it:**

`.compute_cmr_metrics()` gained `periodicity_check = c("abort", "warn")`, defaulting to `"abort"`. **No production call site passes it** — the three `cmr_metrics_*` targets use the abort default, so as merged the build goes red.

If you want the finding published before the remedy is chosen, add `periodicity_check = "warn"` to the three call sites in `R/plan_commodities_mean_reversion.R` (lines ~126–136) for exactly one release. Verified: under warn mode the pipeline computes normally (`vol = 0.2085` for 1m, identical to the current store) and emits the full abort text as a `cli_warn`. It is not silent — a silent pass is the failure mode #738 is about.

I would not leave it in warn mode past the remedy decision. Given the measurement above, the vol error is 1.9% and the σ impact is 0.09%, so there is no risk-management urgency forcing a red build — but the CAGR/Sharpe flattering and the 25% annualisation-span error are real and are what the warning should keep visible.

## Changes NOT made (flagged, per scope)

Three other files carry the same wrong premise. Two other agents are in them right now, so I have described rather than edited:

1. **`R/plan_strategy_names.R`** — `cmr` (position 11) still declares `frequency = "monthly"`, `ann_factor = 12L`, which feeds `bt.strategy.frequency` / `bt.strategy.ann_factor` via `.cmr_register_runs()`. Flagged by #720 as well and still open. Note `R/plan_leaderboard.R`'s `STRATEGY_OBS_ANN_FACTOR` correctly says 252 for CMR — **the two tables disagree with each other**, which is its own `fail-loud-not-null` defect (no gate reconciles them).
2. **`.cmr_register_runs()`** (`R/plan_commodities_mean_reversion.R`, in my file but outside this change's remit) hardcodes `hd_record_stability_metrics(..., w = 36L, ann_factor = 12L)` with the comment "Monthly series". Same root cause as #717, still wrong after #720.
3. **#719 Layer 3** specifies the median-gap approach that #738 shows is insufficient. That specification should be amended to the consistency form. The two constants and the check body here are self-contained and lift cleanly into a shared helper — the only CMR-specific parts are the function name and the message prefix.

Also flagged for a separate issue: **`n_long = 10, n_short = 10` against a 13-name pre-2000 universe** (see Half 1). No periodicity remedy fixes that.

## Verification

`scripts/verify.sh`, full run, foreground:

```
PASS: _targets.R parses
PASS: docs/_targets.R parses
PASS: docs/_targets.R is structurally valid (tar_validate, 7.50s)
PASS: dashboard freshness

SKIP count this run: 0

SKIP count this run: 15

```

Baselines per `.claude/CLAUDE.md` (root `failed=3 skipped=0`, package `failed=1 skipped=15`) hold exactly.

New tests: `tests/testthat/test-cmr-periodicity-consistency.R` — 13 `test_that` blocks, 26 assertions, **4 `expect_snapshot(error = TRUE, ...)`** (31%, meeting `snapshot-test-policy`'s 30% floor for 9+ blocks), `testthat::local_edition(3)` at the top. Coverage: clean daily passes; 9/11-sized closure + weekends pass; clean monthly passes; two isolated outages pass, three fire; the CMR defect shape fires; the too-short direction fires; the classifier/tolerance tables cannot drift apart; warn mode warns and is not silent; band overlap is harmless. `tests/testthat/test-cmr-units.R` still passes unchanged (28 assertions) — its monthly toy fixtures at `ann_factor = 12L` are consistent and are not affected.

`tar_make()` was **not** run.

Refs #738, #717, #720, #719, #635, #736

### DIFF
--- a/R/plan_commodities_mean_reversion.R
+++ b/R/plan_commodities_mean_reversion.R
@@ -32,6 +32,13 @@
 #   own date column still mixes daily and monthly-stamped observations, so
 #   ann_factor = 252 over-annualises the vol/cagr contribution of every
 #   FRED-only month-start row. #724 fixes the crash; it does not fix that.
+# Mixed frequency (#738): MEASURED, not fixed. cmr_portfolio_1m/3m/6m run at
+#   12 obs/year before 2000-01 (13 FRED monthly indexes only -- Yahoo history
+#   starts 2000) and ~255/year after; ~1.4% of observations are monthly-spaced
+#   and the whole series is declared daily. .assert_cmr_ann_factor()'s
+#   consistency check (#738) now DETECTS this; choosing the remedy (truncate
+#   to 2000+, resample to monthly, or segment and report separately) is an
+#   open decision on #738 and is deliberately NOT made here.
 
 plan_commodities_mean_reversion <- function() {
   list(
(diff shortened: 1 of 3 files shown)


JohnGavin:
Superseded by #750, now merged — it carries this PR's commit (`236e945`) plus the
staging change the human approved.

The guard in this PR is correct and unaltered. The only difference is that the three
production call sites now pass `periodicity_check = "warn"`, so the finding is loud in
every build log rather than turning `main` red while the underlying data question is
still open. `.compute_cmr_metrics()`'s own default remains `"abort"`, so a new call
site that omits the argument inherits the strict behaviour — leniency has to be opted
into, never inherited.

Observed on the real store after merge: **777 targets inspected, 0 errored**,
`build.sh` PASS, and the warning fires on all three lookbacks with its full text:

```
CMR 3m: the observation spacing is NOT consistent with a single declared
periodicity (ann_factor 252, daily).
92 of 6847 gaps (1.344%) fall outside the daily band [1, 10] calendar days;
  allowance is 7.
Observed gap bands: daily: 6755; monthly: 92.
Out-of-band gaps range 28-31 calendar days, spanning 1992-06-01..2000-01-01.
The median gap (1 calendar days) agrees with the declared factor, which is why
  the #720 median-gap check passes -- a median cannot see a minority at a
  different frequency.
```

That last line is the part worth keeping: it states its own predecessor's blind spot
rather than leaving a reader to rediscover it.

## Correction to this PR's report

Two items it flagged as follow-ups are **not defects** — I checked both:

- `R/plan_strategy_names.R` does **not** declare `cmr` as monthly/12. Line 91 reads
  *"#717: cmr (position 11) is daily, not monthly"*, and evaluating the table gives
  `cmr | daily | 252`. Fixed by #736.
- `.cmr_register_runs()` does **not** pass `ann_factor = 12L`. The live call passes
  `w = 252L, ann_factor = 252L`; the `12L` references nearby are comments describing
  the *monthly* strategies.

Recording this so the false leads do not get actioned later.

Closing in favour of #750. Nothing is lost — the guard is in `main`.

Refs #738, #750.

Issue #751, opened 2026-08-24
CMR holds untradeable IMF indexes, ranks the same commodity against itself, and its '3m' lookback is 3 days for 96% of the data — three defects worse than the frequency break


JohnGavin:
Commissioned research into how mixed-frequency series merging is actually done, rather
than inventing an approach. The literature answer is deflationary and clear. **But the
research surfaced three defects more serious than the frequency break, and they change
what the right remedy is.** I have verified all three directly against the code; they
are not taken on the researcher's word.

---

### 1. The pre-2000 names are not investable instruments

The 13 pre-2000 series are **IMF Primary Commodity Price System indexes served via
FRED** — `POILWTIUSDM`, `PCOPPUSDM`, `PWHEAMTUSDM` etc.
(`scripts/fetch_commodities.R:39-71`). These are *statistical price indexes*. You
cannot buy one, and you certainly cannot short one.

Every position CMR takes before 2000 is in something that is not a security. This is
not a frequency problem — it is a "the backtest holds positions in things that do not
trade" problem, and no resampling fixes it.

The IMF further documents that monthly values are **time-averages of weekly or daily
prices**. Averaged series carry spurious first-order serial correlation by construction
([Working 1960](https://cowles.yale.edu/sites/default/files/2022-08/cowles-p159.pdf)),
which an autocorrelation-based signal — and cross-sectional mean reversion is one —
reads as signal.

### 2. The cross-section ranks the same commodity against itself

Verified in `scripts/fetch_commodities.R`. The universe contains **three
representations of WTI crude**: `POILWTIUSDM` (IMF monthly index, line 42), `CL=F`
(futures, line 110), `USO` (ETF, line 130). Also gold twice (`GC=F` + `GLD`), silver
twice (`SI=F` + `SLV`), and `DBC` / `PDBC` / `DBA` — **baskets whose constituents are
already held as separate names**.

The researcher's estimate is ~12–14 distinct underlying exposures behind a nominal 23.

The damaging interaction — flagged as **inference**, though it follows directly from
[Lo & MacKinlay (1990)](https://www.sciencedirect.com/science/article/abs/pii/030440769090098E)
on non-synchronous trading: a smoothed index lags the unsmoothed futures price of the
same commodity. After a sharp move the two disagree, a mean-reversion signal ranks the
laggard long and the leader short, and **both legs are the same underlying**. That is a
stale-price artifact, not mean reversion — and it is unrealisable, because one leg is
not a security.

[...]

PR #752, opened 2026-08-24
fix(cmr): calendar-based lookback window for hd_commodity_mr_signal() (#751 item E)

### LINKED ISSUE #738: CMR's series is monthly before 2000 and daily after, declared daily throughout — and the median-gap guard cannot see it

JohnGavin:
## Summary

Item **E** from #751: makes `hd_commodity_mr_signal()`'s `lookback_months`
argument a **calendar window** (elapsed days), not a row-count window.

Under the previous row-count window, `lookback_months = 3` gave a FRED/IMF
monthly series a genuine 3-month window and a Yahoo daily series a
3-*trading-day* window, and both were then ranked in the same
cross-sectional sort. A 3-month return has ~sqrt(21) times the dispersion
of a 3-day return, so the monthly names occupied the extreme ranks
near-mechanically whenever they printed -- filling both the long and short
buckets. This is verified still live through 2026-03-01 (13 IMF series
still print monthly), not only a pre-2000 artifact.

The file's own header comment previously called this "cosmetic-only" per
#720. That classification was wrong for this mixed-cadence universe and is
corrected in this PR.

## What changed

- `hd_commodity_mr_signal()` (`packages/historicaldata/R/commodities_mean_reversion.R`)
  now computes the cumulative lookback return with `slider::slide_index_dbl()`
  keyed on `date`, with `.before = lubridate::days(round(lookback_months * 365.2425/12))`
  -- a fixed calendar-day window, so `lookback_months = 3` spans the same
  ~91-day economic horizon for every series regardless of how often it
  prints.
- `lubridate::days()` (fixed days) is used rather than `months()` period
  arithmetic, because `months()` subtraction produces `NA` at day-31
  anchors with no equivalent day N months prior (the same gotcha
  `strategy_mom_prepeak.R`'s `%m-%` workaround exists for).
- **Completeness floor (fail-loud-not-null.md):** `slider`'s own
  `.complete = TRUE` only checks that a window's start boundary falls
  inside a series' observed date range -- it does not check how many
  observations actually landed inside the window. A window can span the
  right number of days while crossing a data gap and containing almost no
  real observations. Added a per-series minimum-observation floor:
  `max(.HD_MR_MIN_OBS_FLOOR = 2, ceiling(.HD_MR_MIN_OBS_FRACTION = 0.5 *
  window_days / series' own median gap))`. Windows failing this floor
  produce `NA`, same as an insufficient-history window always has. See the
  new `.HD_MR_MEAN_DAYS_PER_MONTH` / `.HD_MR_MIN_OBS_FRACTION` /
  `.HD_MR_MIN_OBS_FLOOR` roxygen blocks for the full reasoning behind each
  constant.
- File header and function roxygen updated to describe the new semantics
  and correct the "cosmetic-only" mischaracterization; the `monthly_ret`
  column is **not** renamed (see "What I deliberately did not do" below).
- 3 new tests replace the old row-count-based look-ahead test (which
  hand-computed expected signals from row indices -- a description of the
  *old* implementation, not a property of the function) and add explicit
  coverage for the fix and the new floor:
  1. Look-ahead guard rewritten as a **property test**: perturbing every
     return at/after a cutoff date must not change any signal computed for
     a date strictly before the cutoff -- true regardless of internal
     window implementation.
  2. Directly demonstrates the #751 defect and its fix: a daily series and
     a monthly series with the same `lookback_months = 3` now capture
     ~63 vs ~3 observations respectively (previously both would have been
     exactly 3).
  3. Directly demonstrates the new completeness floor: a dense daily
     series with an internal 70-day gap has its post-gap window correctly
     excluded (not silently accepted as complete) until enough dense data
     has re-accumulated.

## What I deliberately did not do (per dispatch scope)

- **Did not touch universe composition** -- no series dropped, no ETF/futures
  dedup, `n_long`/`n_short` unchanged. Those are separate open decisions on
  #751.
- **Did not truncate to 2000+.** Also a separate #751 decision.
- **Did not touch the #738 periodicity guard** (`.assert_cmr_ann_factor()`,
  staged at `periodicity_check = "warn"` per #750) -- untouched, still
  fires on the same underlying mixed-frequency condition (see measurement
  below). This PR does not resolve #738.
- **Did not rename `monthly_ret`.** It is misleadingly named (holds a daily
  return for 96% of rows), but renaming ripples through every caller in
  this file and the plan file, which would fold an unrelated mechanical
  change into this one and violate "one change per verification run." Fixed
  the misleading comment/roxygen instead; renaming is left as a follow-up if
  wanted.

## Verification

**OBSERVED** (all commands run this session):

1. `scripts/verify.sh` (full run, foreground) -- **PASS**.
   - Root suite: `total=755 failed=3 skipped=0` -- matches the documented
     baseline exactly (3 known failures: `test-crypto-momentum.R`,
     `test-qa-summary-deps.R`, `test-stock-backtest.R`). Total count differs
     from the `.claude/CLAUDE.md`-documented 739 due to unrelated commits
     since that baseline was written; the failure *signatures* match
     exactly, which is what `scripts/verify.sh` actually checks.
   - Package suite: `total=715 failed=1 skipped=15` -- matches the
     documented baseline exactly (1 known failure:
     `test-mom-prepeak-portfolio.R`; skip count 15 = 4 `{alphavantager}`-absent
     + 11 `HD_TEST_LIVE`-gated).
   - `parse("_targets.R")`, `docs/_targets.R` parse + `tar_validate`,
     testthat edition 3 active, dashboard freshness -- all PASS.

2. `packages/historicaldata/tests/testthat/test-commodities-mean-reversion.R`
   in isolation -- 23/23 assertions pass.

3. **Before/after live-store comparison** (read-only `tar_read()` against
   `docs/_targets`, **no** `tar_make()` run; new function evaluated directly
   against the same `commodities_returns`/`daily_rf` the live store already
   has):

   | lookback | sharpe before | sharpe after | cagr before | cagr after | vol before | vol after | max_dd before | max_dd after |
   |---|---|---|---|---|---|---|---|---|
   | 1m | -1.86  | -0.438 | -0.369 | -0.0686 | 0.208 | 0.200 | -1.000 | -0.866 |
   | 3m | -1.23  | -0.029 | -0.257 |  0.0126 | 0.225 | 0.228 | -1.000 | -0.722 |
   | 6m | -0.747 |  0.14  | -0.152 |  0.0508 | 0.229 | 0.228 | -0.996 | -0.718 |

   Sharpe moves from deeply negative to roughly flat/slightly positive
   across all three lookbacks. This is consistent with #751's own
   prediction ("most of it does not [survive]") -- reported plainly, not
   softened. The `.compute_cmr_metrics()` call used `periodicity_check =
   "warn"`, matching the current staged production config (#750); the #738
   periodicity warning still fires (39-93 of ~6700-6800 gaps out of band per
   lookback) since this PR does not touch that guard.

**PREDICTED, not run** (explicitly out of scope -- `scripts/build.sh` holds
a lock and `tar_make()` was not run): that a full `tar_make()` on this
branch reproduces the "after" figures above through the real pipeline. The
live-store read above exercises the identical code path
(`hd_commodity_mr_signal` -> `hd_commodity_mr_portfolio` ->
`.compute_cmr_metrics`) against the same live data, so this is expected but
not directly observed via `tar_make()` itself.

## Refs

Refs #751, #738, #750, #720, #717

### DIFF
--- a/packages/historicaldata/R/commodities_mean_reversion.R
+++ b/packages/historicaldata/R/commodities_mean_reversion.R
@@ -14,50 +14,149 @@
 #   is daily-dominated (ann_factor=252), not monthly (ann_factor=12) despite
 #   this file's "monthly"/lookback_months naming throughout -- see
 #   R/plan_commodities_mean_reversion.R's file header for the source
-#   evidence. `lookback_months` here counts unique dates in the merged
-#   universe (one row per date), not calendar months; cosmetic-only fix,
-#   left un-renamed per #720's note (shared with commodities_momentum.R).
+#   evidence. The `monthly_ret`/`lookback_months` naming is left un-renamed
+#   per #720's note (shared with commodities_momentum.R); `monthly_ret` holds
+#   a per-observation return whose actual period is daily for 96% of the
+#   merged universe's rows (Yahoo futures/ETF series) and monthly for the
+#   rest (FRED/IMF indexes) -- the name is a historical label, not a
+#   guarantee about the row's cadence.
+# #751: `lookback_months` was PREVIOUSLY a row-count window (the `lb`-th
+#   previous ROW, regardless of its date) -- correctly "cosmetic-only" per
+#   #720 only for a single-frequency series, but WRONG for this mixed-cadence
+#   universe: `lookback_months = 3` gave a FRED monthly series a genuine
+#   3-month window and a Yahoo daily series a 3-*trading-day* window, and
+#   both were then ranked in the same cross-sectional sort. A 3-month return
+#   has ~sqrt(21) times the dispersion of a 3-day return, so the monthly
+#   names occupied the extreme ranks near-mechanically whenever they printed.
+#   Verified still live through 2026-03-01 (13 IMF series still print
+#   monthly), not just a pre-2000 artifact. hd_commodity_mr_signal() now
+#   uses a CALENDAR window (elapsed days), so `lookback_months = 3` means the
+#   same economic horizon for every series regardless of how often it prints
+#   -- see that function's roxygen for the window/floor design.
 # Universe re-uses commodities_returns from plan_commodities_momentum.R.
 
 
+#' Average Gregorian calendar month length, in days (365.2425 / 12)
+#'
+#' Used to convert \code{lookback_months} into a fixed number of CALENDAR
+#' DAYS for \code{\link{hd_commodity_mr_signal}}'s sliding window, rather
+#' than using \code{lubridate::months()} period arithmetic directly.
+#' \code{months()} subtraction produces \code{NA} whenever the anchor date
+#' has no equivalent day N months prior (e.g. day 31 minus 3 months, when
+#' the target month has fewer than 31 days) -- see
+#' \code{strategy_mom_prepeak.R}'s \code{lubridate::`%m-%`} workaround for
+#' the same underlying gotcha. A fixed-day window sidesteps this entirely:
+#' \code{date - lubridate::days(n)} is always defined. 30.436875 (=
+#' 365.2425 / 12) is the standard Gregorian mean month length; using it
+#' rather than a round number (30) keeps a 12-month window close to 365
+#' days rather than drifting short by ~4 days a year.
+#'
+#' @noRd
+.HD_MR_MEAN_DAYS_PER_MONTH <- 365.2425 / 12
+
+#' Minimum window-fill fraction required for a calendar window to be treated
+#' as complete
+#'
+#' A calendar window can span the right number of days while containing
+#' almost no observations, if it happens to cross a data gap (a stale/thin
+#' patch of a series, or -- for this mixed-cadence universe -- a series that
+#' is far sparser than a typical window expects). \code{slider}'s own
+#' \code{.complete = TRUE} only checks that the window's start boundary
+#' falls inside the series' observed date RANGE; it does not check how many
+#' observations actually fall inside the window.
+#'
+#' This fraction is applied against \code{window_days / median_gap_days} --
+#' the observation count a window of this length would contain if the
+#' series kept up its own typical (median) spacing. 0.5 tolerates normal
+#' weekend/holiday sparsity in a daily series and the +/-1 rounding a
+#' ~30.44-day-average month produces against real (non-uniform) monthly
+#' print dates, while still rejecting a window degraded by an outage/gap
+#' materially larger than the series' own norm. It is deliberately not
+#' closer to 1: a strict 90-100% fill would reject completely ordinary daily
+#' windows that happen to contain an extra holiday cluster.
+#'
+#' @noRd
+.HD_MR_MIN_OBS_FRACTION <- 0.5
+
+#' Absolute floor on observations inside a calendar window, regardless of
+#' cadence
+#'
+#' A "cumulative return over a lookback window" is a product over multiple
+#' observations; a window containing only the observation at \code{t} itself
+#' is not a lookback return at all, it is that single observation's own
+#' return relabelled as a longer-horizon one. 2 is the smallest count for
+#' which the product spans more than one observation. This floor matters
+#' when a series' own cadence cannot be estimated (fewer than 2 dated
+#' observations) or is so sparse that \code{.HD_MR_MIN_OBS_FRACTION} of the
+#' cadence-implied count would round below it.
+#'
+#' @noRd
+.HD_MR_MIN_OBS_FLOOR <- 2L
+
 #' Commodity Mean Reversion Signal
 #'
-#' Compute a monthly mean-reversion signal for a commodity universe.
+#' Compute a mean-reversion signal for a commodity universe over a fixed
+#' CALENDAR window (elapsed days), not a fixed number of rows.
 #' The signal is the *negative* of the lookback-period return: commodities
 #' that have fallen the most receive the highest (most positive) signal,
 #' while those that have risen the most receive the lowest (most negative)
 #' signal.
 #'
-#' Look-ahead safety: the signal at month \code{t} is constructed from
-#' cumulative returns over months \code{(t - lookback_months)} through
-#' \code{(t - 1)}.  The one-period lag is enforced via
-#' \code{dplyr::lag(cumret)} before the negation, so no return at or after
-#' month \code{t} ever enters signal formation.
+#' Look-ahead safety: the signal at date \code{t} is constructed from
+#' cumulative returns over the \code{lookback_months} calendar months
+#' immediately preceding \code{t}, ending at \code{t - 1}'s observation. The
+#' one-period lag is enforced via \code{dplyr::lag(cumret)} before the
+#' negation, so no return at or after date \code{t} ever enters signal
+#' formation.
 #'
-#' @param returns_tbl Tibble with columns \code{date} (Date, month-end),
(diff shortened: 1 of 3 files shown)


JohnGavin:
Verified against the live store. Three things to correct or record before this merges.

## 1. The pipeline did not pick this change up — filed as #753

Built this branch twice against the same store:

| run | CMR sharpe | months | verdict |
|---|---:|---:|---|
| `scripts/build.sh` (normal) | **−0.747** — identical to pre-fix | 6789 | `777 targets, 0 errored, PASS` |
| after `tar_invalidate(starts_with("cmr_"))` | **+0.14** | 6669 | `777 targets, 0 errored, PASS` |

A change under `packages/*/R/` does not invalidate the targets that call it. The first
build skipped all 15 `cmr_*` targets and served the old function's results while
reporting PASS. Filed as **#753**; it is a more serious defect than the one this PR
fixes, because it defeats the verification protocol that catches everything else.

Nothing wrong with this PR — but "I built it and the numbers didn't move" would have
been the wrong conclusion, and it is the conclusion the tooling hands you.

## 2. The direction contradicts #751's prediction — it does not confirm it

The PR describes the swing as *"consistent with #751's prediction that most of CMR's
previously measured performance was a mixed-horizon-ranking artifact."*

That reading is backwards. #751's mechanism was that the stale-price and mixed-horizon
effects generate **spurious profit** — going long the stale index and short the futures
of the same commodity, with both legs winning. If that were the dominant effect,
removing it would make measured performance **worse**.

Observed: Sharpe **−0.747 → +0.14**, an improvement of ~0.89. CAGR −15.2% → +5.1%.
Max drawdown −99.6% → −71.8%. Every headline number improved.

So the artifact was destroying measured performance, not manufacturing it. The
plausible reading — and I flag this as **inference**, not measurement — is that ranking
3-month IMF returns against 3-day futures returns put the monthly names in the extreme
buckets near-mechanically, and those positions were systematically wrong rather than
systematically right. That is a different mechanism from the one #751 hypothesised, and
it should be recorded as such rather than filed as a confirmation.

**This does not make CMR a good strategy.** From the same rebuilt store:

| | |
|---|---:|
| sharpe | 0.140 |
| **deflated_sharpe** | **−0.057** |
| detection_min_n_years | 315.4 |
| detection_underpowered | TRUE |

After adjusting for the multiplicity we actually incur, the edge is **negative**, and
detecting a 0.14 Sharpe would need 315 years against the ~26 available. The correct
summary is not "CMR works after all" but "CMR's measured loss was substantially an
artifact, and what remains is indistinguishable from zero."

## 3. Two changes are bundled, so the swing is not attributable

This PR makes the window calendar-based **and** adds a completeness floor
(`.HD_MR_MIN_OBS_FRACTION` / `.HD_MR_MIN_OBS_FLOOR`) that rejects windows crossing a
data gap. The floor drops **120 observations** (6789 → 6669), so the sample changed too.

The −0.747 → +0.14 move therefore has two possible causes in unknown proportion.
`verification-before-completion`'s "One Change Per Verification Run" exists for exactly
this.

I am not asking for a split, because a calendar window cannot be defined without
defining what "complete" means for it — the floor is arguably part of the same change
rather than a second one. But the attribution gap should be on the record, and if the
split ever matters, the control is this branch with the floor set to accept any
non-empty window.

Refs #751, #753.

Issue #753, opened 2026-08-24
Changes under packages/*/R/ do not invalidate dependent targets — tar_make() skips them, build.sh reports PASS, and tar_read() serves the old function's results


JohnGavin:
A change to any file under `packages/historicaldata/R/` does **not** invalidate the
targets that call the changed function. `tar_make()` skips them, `build.sh` reports
`PASS`, and `tar_read()` serves values computed by the **previous** version of the code.

No error. No warning. The build is green and the numbers are stale.

## Reproduced today, on PR #752

#752 rewrites `hd_commodity_mr_signal()` in
`packages/historicaldata/R/commodities_mean_reversion.R` — the CMR lookback window.
Built against the live store twice, same branch, same store:

| run | CMR Full-Period sharpe | months | build verdict |
|---|---:|---:|---|
| `scripts/build.sh` (normal) | **−0.747** — identical to pre-fix | 6789 | `777 targets, 0 errored, PASS` |
| `tar_invalidate(starts_with("cmr_"))` then `scripts/build.sh` | **+0.14** | 6669 | `777 targets, 0 errored, PASS` |

The first run skipped every `cmr_*` target (`720 skipped`) and served cached results.
Only the forced invalidation revealed that the change had any effect at all.

## Why

`docs/_targets.R:25` loads the package with:

```
pkgload::load_all(here::here("packages/historicaldata"), quiet = TRUE)
```

This is the correct mechanism — `historicaldata` is not installed in the nix
environment — and the functions it defines *are* live in the session. But `targets`
builds its dependency graph from objects it can see in the pipeline's environment at
parse time and from files declared as `tar_target(..., format = "file")`. A function
that arrives via `load_all()` inside the script body is not tracked as a dependency of
the targets that call it, so changing its body does not change any target's hash.

Root `R/plan_*.R` files are **not** affected — they are sourced into the pipeline
environment and their functions are tracked normally. This is specifically a
**package-source** blind spot.

## Why this is worse than the bugs it hides

This is the same lesson this repo has now learned four times, one level further up:

- **#690** — `verify.sh` was checking the wrong pipeline
- **#691** — `error = "continue"` let errored targets exit 0
- **#730** — `check_pipeline_errors.R` failed open on a truncated meta read
- **this** — package edits produce no rebuild at all

Each time, *"the pipeline is green"* meant *"nothing changed"*. Here it means it most
literally: nothing changed, because nothing was rebuilt.

[...]

Issue #730, opened 2026-08-23
check_pipeline_errors.R fails open on an unreadable store: tar_make() printed 'errored pipeline', the checker printed 'PASS: no errored targets'


JohnGavin:
`scripts/check_pipeline_errors.R` exists to catch the case where `tar_make()` exits 0
while targets have errored (`error = "continue"` in `docs/_targets.R`). It has a
failure mode of its own: **when the store's metadata file is unreadable, it reports a
clean pipeline.**

## Observed

Two `tar_make()` runs overlapped against one store — my error, and the session-init
banner warns about exactly it. The result was a single interleaved line in
`docs/_targets/meta/meta`: the `leaderboard` record written twice, concatenated
(35 fields where 18 are expected; the two halves differing only in a timing field,
`1.693` vs `1.692`).

What each layer then reported:

| layer | output |
|---|---|
| `tar_make()` | `✖ qa_summary errored` — `Error storing output: cannot open the connection` |
| `tar_make()` | `✖ errored pipeline [31.4s, 39 completed, 730 skipped]` |
| `check_pipeline_errors.R` | `Targets inspected: 762` / `Targets errored: 0` / **`PASS: no errored targets`** |
| `scripts/build.sh` | **`PASS (tar_make() clean, no errored targets, dashboard freshness clean)`** |

The pipeline said *errored* on the line above, and the checker built to catch that
said *clean*.

## Mechanism

`tar_meta()` reads the meta file with `data.table::fread()`, which hit the malformed
line and stopped:

```
Stopped early on line 1023. Expected 18 fields but found 35.
```

That is a **warning**, not an error. `fread()` returned the 762 rows it managed to
parse, `tar_meta()` returned them, the checker found no `error` field set among them,
and reported PASS. The nine missing targets — including `leaderboard` and the errored
`qa_summary` — were simply not in the frame it inspected.

So the checker is not wrong about what it saw. It has no way to know it saw a
truncated view, because nothing compares the row count against anything.

## Why this matters beyond the concurrency mistake

The concurrency was avoidable and mine. The fail-open is not incidental to it: **any**
cause of an unreadable or partial meta file produces the same silent PASS — an
interrupted write, a full disk, a killed process mid-flush, an editor touching the
file. The checker's whole purpose is to be the layer that does not take `exit 0` at
face value, and it takes `fread` at face value instead.

It is also, structurally, the same defect this repo has now fixed four times: a
[...]

PR #757, opened 2026-08-25
fix(pipeline): track packages/historicaldata/R as a targets dependency (#753)

JohnGavin:
## Summary

A change to any file under `packages/historicaldata/R/` did not invalidate the
targets that call the changed function — `tar_make()` skipped them,
`scripts/build.sh` reported `PASS`, and `tar_read()` served the previous
version's value. Reproduced twice on PR #752's CMR lookback fix (same branch,
same store: `-0.747` on a normal `build.sh` run, `+0.14` only after a forced
`tar_invalidate()`).

Root cause: `pkgload::load_all()` (`docs/_targets.R:25`) is how
`historicaldata` gets loaded — functions arriving that way are not tracked as
target dependencies the way root `R/plan_*.R` sourced functions are.

## Fix — three parts (issue #753's options 2 + 4, refined after empirical testing)

1. **`tar_option_set(imports = "historicaldata")`** in `docs/_targets.R` —
   `targets`' own built-in mechanism for this exact problem. Tracks every
   **bare-name** package call (`hd_fn()`) automatically, zero per-target
   edits. Confirmed empirically (scratch pipeline) to match PR #752's exact
   call shape. **Does not** cover `historicaldata::fn()` **namespaced**
   calls — `targets`' own documentation states this is a
   `codetools::findGlobals()` limitation, also confirmed empirically. 22
   files / ~135 lines in this repo's `R/` use that form.

2. **`pkg_source_files` (`format = "file"`) + `pkg_source_digest` targets**
   — a file-hash-tracked digest of package source, independent of
   `pkgload::load_all()` (so immune to the defect it's guarding against).
   Any target can opt in by referencing `pkg_source_digest` in its own
   command. `qa_pkg_source_tracked` (S22, `R/plan_qa_gates.R`) guards this
   mechanism itself from going silently vacuous (e.g. the package directory
   being renamed and `pkg_source_files` quietly returning `character(0)`).

3. **`scripts/check_pkg_staleness.R`**, run unconditionally as `build.sh`
   Step 2c — the real cross-run staleness detector. Discovers every target
   defined in a file that mentions `historicaldata::` (no hand-maintained
   registry — see `.cps_discover_consuming_targets()`), and fails the build
   with `[STALE-PKG] <name>` if any of them was **skipped** in a run where
   the package source changed.

### Why part 3 isn't a `tar_target()` gate (design deviation, explained)

The dispatch asked for an in-pipeline S-gate matching S19–S21's shape.
Empirically this is impossible: `tar_meta()`/`tar_progress()` are explicitly
unsupported when called from within a target of the pipeline currently
running —

```
Error: target gate attempted to run targets::tar_meta() to during a
pipeline, which is unsupported except when format is "file" and
repository is "local", or if you are reading from a data store that does
not belong to the current pipeline. ...
```

— confirmed with a throwaway scratch pipeline before writing any real code
(see `scripts/check_pkg_staleness.R`'s header comment for the full
transcript). `check_pkg_staleness.R` therefore follows the SAME pattern this
repo already trusts for this class of check: `check_pipeline_errors.R` and
`check_dashboard_freshness.R` are both separate post-build scripts, run as
`build.sh` steps, for exactly this reason (`tar_meta()` calls from the
pipeline's own target bodies are forbidden elsewhere too — see
`.make_store_reader()` in `R/utils_validation.R`).

### Design constraint: false PASS is worse than false FAIL

- `pkg_source_files` intentionally includes `DESCRIPTION` and `NAMESPACE`,
  not just `*.R` — both affect what `pkgload::load_all()` puts in the
  namespace / what `historicaldata::fn()` can resolve to.
- The staleness check combines `tar_progress()` (this-run outcome) with
  `tar_meta()` (build time) specifically to avoid a **false-positive** trap
  on a fresh/full rebuild, where DAG scheduling order is arbitrary and a
  naive timestamp-only comparison would misfire — verified with a dedicated
  "fresh build → PASS" test.
- The registry discovery over-approximates **per file**, not per call site
  — noisier, not quieter.
- `scripts/build.sh` gates Step 2.5 (metadata snapshot) and Step 4
  (`--render`) on the new check too — a stale package-derived value must
  not get committed as a "verified" build time or published to HTML.

### Known, honest gaps (not silently implied as covered)

- The `check_pkg_staleness.R` registry is namespaced-call-based. An indirect
  call (`do.call()`, a stored function reference) that `codetools` also
  can't see via `imports=` is not caught by either mechanism. Nothing in
  this repo currently does that for `historicaldata::`, but it's a real gap.
- Root `_targets.R` (the small sandbox pipeline) doesn't use
  `pkgload::load_all()` at all — confirmed via grep — so it's out of scope,
  not silently unfixed.

## Verification

**`scripts/verify.sh` (full, foreground):** PASS. Root suite: 767 total, 3
failures — exactly the known baseline (unchanged). Package suite: 715
total, 1 failure, 15 skips — exactly the known baseline. 30 new tests (12 +
18) all green, 0 new failures.

**Reproduced the #753 failure shape in a throwaway store** (both as
automated tests in `tests/testthat/test-check-pkg-staleness.R` and as an ad
hoc CMR-shaped demonstration):

```
############ BEFORE: fresh build, everything completed together ############
PASS: no known package-consuming target is stale relative to pkg_source_digest (1 checked)
EXIT STATUS: 0

############ AFTER: package source changed, cmr_signals_1m SKIPPED (repro of #753) ############
FAIL: 1 target(s) were skipped in a run where the package source changed:
  [STALE-PKG] cmr_signals_1m -- last built ... (progress this run: skipped), predates pkg_source_digest (...)
EXIT STATUS: 1
```

**Smoke-built the three new targets against the real `docs/_targets.R`**
(throwaway store, not the real one):

```
✔ pkg_source_files completed [134ms, 586.36 kB]
✔ pkg_source_digest completed [4ms, 85 B]
✔ qa_pkg_source_tracked: S22 passed (64 package source file(s) tracked, digest fd86f711cfeb)
✔ ended pipeline [2.7s, 3 completed, 0 skipped]
```

**Did not** run `tar_make()` against the real store at
`docs/_targets` — `scripts/build.sh` holds a lock refusing worktree builds by
design; all builds above used throwaway `withr::local_tempdir()` /
`tempfile()` stores. The orchestrator will build this branch against the
real store to confirm end to end.

Refs #753, #752, #690, #691, #730.


JohnGavin:
Verified against the real store. **The mechanism is sound and the gate genuinely fires
— but its target attribution is file-level, which makes the build permanently red on
false positives.** Not merging as-is; the fix looks small.

## What works

- `pkg_source_files` / `pkg_source_digest` — file-hash tracking is the right mechanism
  and is immune to the `load_all()` blind spot.
- The empirical finding that an in-pipeline `tar_target()` gate is **impossible** here
  (`tar_meta()` is unsupported from within a target of the currently-running pipeline)
  is correct, was confirmed against the literal error before any code was written, and
  the deviation to a `scripts/` detector matches how `check_pipeline_errors.R` and
  `check_dashboard_freshness.R` already work. Good call, well evidenced.
- Step 2c fires. On the real store it produced **147 `[STALE-PKG]` lines** and
  `=== scripts/build.sh: FAIL ===`. It is not a no-op.

## The defect: file-level attribution

`.cps_discover_consuming_targets()` reads each file under `R/` and, **if the file
contains `historicaldata::` anywhere**, extracts and flags *every* `tar_target()` in
that file.

So a target is flagged for its **neighbours'** package use, not its own.

Worked example from the failing run — `xgb_vs_enet`, flagged as stale:

```r
targets::tar_target(xgb_vs_enet, {
  library(dplyr)
  xgb  <- xgb_drif_portfolio  |> select(ym, xgb_ret  = port_ret)
  enet <- stk_drif_portfolio |> select(ym, enet_ret = port_ret)
  inner_join(xgb, enet, by = "ym") |>
    arrange(ym) |>
    mutate(xgb_cum = cumprod(1 + xgb_ret), ...)
})
```

It calls **no package function at all** — only `dplyr` and two upstream targets. It is
flagged solely because `R/plan_xgb_signal.R` contains 8 `historicaldata::` calls
elsewhere in the file.

## Why this is not merely noisy

There is **no path back to green**. The flagged targets are skipped precisely because
nothing they depend on changed — so they keep predating `pkg_source_digest`, and the
check keeps failing, on every subsequent build, forever. A full forced rebuild would
clear it once, but the same state recurs the next time the package changes, since
`imports` cannot reach namespaced call sites (your own documented limitation, 22 files).

I asked for "a false PASS is worse than a false FAIL" and you correctly chose the
noisier side. That instruction stands — but a gate that is red unconditionally conveys
no information, and this repo has already documented that a check nobody can action
becomes a check nobody reads. The requirement is a gate that is **green when things are
fine and red when they are not**, not one that is always red.

## The fix

Attribute at **target level**: flag a target only if its own `tar_target()` body
contains a package call. The bodies are already delimited — you are parsing
`tar_target(name, {` today; extending to the matching close brace, or using
`parse()`/`getParseData()` rather than line regex, gives per-target attribution
directly. `parse()` is the more robust route and this repo prefers AST inspection over
regex for exactly this reason.

Two things to preserve while fixing:

1. **Keep the conservative direction.** Where attribution is genuinely ambiguous —
   a target calling a *local* helper that itself calls the package — flag it. Prefer
   over-flagging to under-flagging; just make it *provably* over-flagging rather than
   whole-file.
2. **Bare-name calls count too.** `imports = "historicaldata"` covers bare-name calls,
   so those targets *are* auto-invalidated and need not be flagged — but only if the
   `imports` mechanism actually works. Worth stating which population each of the two
   mechanisms covers, so the union is visibly complete.

## Also needed: confirm the steady-state cost

`imports = "historicaldata"` caused a mass invalidation on first build — **89+ targets
dispatched against a normal ~50**, and my verification build exceeded the 600s
foreground ceiling where a warm build is ~30s. A one-time reinvalidation is expected
and fine. What is **not** yet established is whether a *second* consecutive build
returns to ~30s.

If it does, the cost is one-time. If every build re-invalidates, that is a 20× slowdown
on the repo's core loop and `imports` should come out, leaving the digest to carry the
whole mechanism. Please measure it: two consecutive builds, report both durations and
both skip counts.

I could not measure this myself — my own attempt churned the store by switching between
this branch and `main`, which invalidates in both directions.

## Not in question

The problem is real and worth solving properly. #752 demonstrated it: a package fix
showed **zero effect** through a normal build and only appeared after
`tar_invalidate()`. That is the failure this PR exists to prevent, and the digest
approach is the right shape for it.

Refs #753, #752.

JohnGavin:
Addressed both blocking points.

## 1. File-level attribution → target-level attribution

`.cps_discover_consuming_targets()` now parses each `R/*.R` file with
`base::parse()` and walks the resulting AST — no line regex. A target is
flagged only if:

- its own `tar_target()`/`tar_target_raw()` call directly contains a
  `historicaldata::fn()`/`historicaldata:::fn()` call anywhere in its own
  subtree, or
- it calls a local (same-file) helper function that itself, directly or
  transitively through further local helpers, does — the conservative
  "ambiguous case" you asked me to keep, made provably over-flagging rather
  than whole-file.

Verified against the real repo (read-only discovery, no store involved):

```
Total flagged targets: 32  across 20 files   (was 220 across 19 files)
xgb_vs_enet flagged?: FALSE   (was TRUE — this is the fix)
cmr_signals_1m flagged?: FALSE   (bare call — correctly left to imports=)
cmr_registry_run flagged?: TRUE  (real historicaldata:: call site — correctly still caught)
```

`xgb_vs_enet`'s own body — `library(dplyr)` + an `inner_join()` of two
upstream targets — touches no package function and is no longer flagged.
Its actual-culprit sibling in the same file, `xgb_drif_register_runs`
(calls `historicaldata::hd_registry_upsert()`), still is.

Population split stated explicitly in the header comment now: `imports =
"historicaldata"` covers bare-name calls; this script's registry covers
namespaced calls; the two don't overlap by construction, so their union is
the full direct-call surface (indirect calls via `do.call()`/stored
function references remain a named, honest gap for both).

Found and fixed a crash while writing this: R's function-formals pairlists
and omitted call arguments can hold the special missing-argument symbol.
Binding it to a `for (part in list)` loop variable and referencing that
variable raises `"argument is missing, with no default"` — confirmed
empirically, fixed by switching all three AST walkers to index-based
iteration (`parts[[i]]`, never assigned to a bare loop variable first).

New regression tests reproduce the `xgb_vs_enet` shape directly (sibling
targets, one flagged one not, in one file), the direct and transitive
local-helper cases, and the missing-arg crash fixture.

## 2. Steady-state build cost of `imports = "historicaldata"`

Couldn't use the real store (main checkout owns it, per your note). Built a
throwaway store instead, loading the REAL `packages/historicaldata` via
`pkgload::load_all()` (same object population your build hashed) with
`imports = "historicaldata"` set, and 3 targets (a bare call, a namespaced
call, a plain-R target):

```
Build 1: ✔ ended pipeline [593ms, 3 completed, 0 skipped]
Build 2: ✔ skipped pipeline [547ms, 3 skipped]     <- no source changes
Build 3: ✔ skipped pipeline [557ms, 3 skipped]     <- no source changes
```

**OBSERVED**: two consecutive no-op builds both fully skip — 0 targets
rebuilt, stable internal timing across all three runs. This directly
answers the question: the mechanism doesn't re-invalidate on repeat: an
`imports` hash is a pure function of unchanged source text, so build 2's
hash matches build 1's stored hash exactly and everything skips.

**PREDICTED, not observed** (labelling this explicitly, as asked): applying
that mechanism to the real 778-target store, a second real build should
return close to your ~30s baseline. I have not measured the real store's
absolute number and I'm not claiming to — only the qualitative
invalidate-vs-skip behaviour, in a reduced but mechanistically faithful
reproduction. Documented inline in `docs/_targets.R` next to the `imports`
option, with both the observed numbers and the predicted/not-observed
caveat spelled out.

## Verification

`scripts/verify.sh` full suite: PASS. Root suite 775 total (767 + 8 new
assertions), 3 known baseline failures unchanged, 0 new. Package suite
unchanged (baseline 1 failure, 15 skips).

Refs #753, #752.

PR #766, opened 2026-08-26
fix(dashboards): shorten leaderboard caption, real links, dark-mode gate, font controls

JohnGavin:
## Summary

- **Task 1** — leaderboard's Rankings caption was 618 words / 4589 chars. Restructured into a short summary + collapsed `<details>` (nothing deleted), converted dead markdown-style `[text](url)` links to real `<a href>` tags via a new `hd_link()`/`hd_caption()` pair in `docs/vignette_utils.R`. Investigated the reported "unclosed caption / 41,400 chars swallowed" claim thoroughly — could **not** reproduce it (xml2-based well-formedness audit of all 78 captions across all 15 dashboards on unmodified `main`: all well-formed). Found a related *latent* risk in `falsification.qmd`'s `fals_dt()` helper instead (relies on hand-typed `&lt;`/`&gt;` escaping); left unchanged pending a working render to verify a fix, flagged for follow-up, and added a permanent regression test instead.
- **Task 2** — audited caption length across all 15 dashboards; no malformed markup anywhere. `hd_dt()` now routes every caption through `hd_caption()`, so any long caption anywhere gets the same short+details treatment automatically.
- **Task 3** — base font size raised two steps (1.1em → 1.3em), moved to a `--base-font-size` CSS custom property.
- **Task 4** — font +/− buttons added next to Quarto's existing dark/light toggle (which needed only an `aria-label`). Persisted via `localStorage`, keyboard/screen-reader accessible.
- **Task 5** — every dashboard now declares `<meta name="color-scheme" content="dark">` + CSS `color-scheme: dark` via a shared `docs/_includes/color-scheme-meta.html`, kept in sync with the live theme at runtime via a `MutationObserver`. Wired `check_dashboard_color_scheme.sh` into `docs/_quarto.yml` post-render.

## Verification

- `scripts/verify.sh`: **PASS** — root suite 3/3 known baseline failures (no new regressions), package suite 1/1 known baseline failure, 28 new assertions across two new test files all pass.
- Re-rendered 12 of 15 dashboards fresh via `quarto render`; all pass `check_dashboard_color_scheme.sh`. Directly patched the 7 dashboard HTML files with no `.qmd` source.
- A full project-level `quarto render docs` confirms the post-render hooks fire correctly and gate on exactly 3 pre-existing, unrelated broken pages (see below).

## Known blockers (pre-existing, out of scope, confirmed via `git stash` A/B against unmodified `main`)

`leaderboard.qmd`, `bdbb-sol.qmd`, and `macro-defense-rotation.qmd` cannot render in this worktree:
- No local `_targets` store (worktrees don't carry one).
- `inst/extdata/vignettes/leaderboard.rds` is stale — missing ~14 columns the current `leaderboard.qmd` depends on (`detection_underpowered`, `credible`, `gross_convention`, etc.). Pinned as a documented "KNOWN ISSUE" test in `tests/testthat/test-hd-caption.R` so a future fixture regen shows up as a passing-test diff.
- `bdbb-sol.qmd` / `macro-defense-rotation.qmd` hit a targets-store path-duplication bug (`docs/docs/_targets not found`) in their non-`safe_tar_read()` calls.

Source fixes for all three are committed and unit-verified regardless; their `.html` artifacts remain stale pending a working render (main checkout with a live store, or the fixture fix).

Also: `api-historicaldata.html` is regenerated by `scripts/regen_api_context.sh` (out of scope for this dispatch) — that script will need updating to keep the color-scheme fix on the next regen.

## Test plan

- [x] `scripts/verify.sh` full suite passes
- [x] `check_dashboard_color_scheme.sh docs` — 20/23 pass (3 pre-existing blockers documented above)
- [x] New `hd_caption()`/`hd_link()` unit tests (well-formedness, boundary lengths, real-link rendering, escaping)
- [x] New dashboard-wide caption well-formedness regression test
- [ ] Manual visual QA of dark/light toggle + font buttons in a real browser (not possible from this environment — no headless browser in the nix env)


JohnGavin:
Superseded by #767, now merged — it carries this PR's two commits plus the wiring fix.

The caption restructure, font property, controls, tests and the project-level `color-scheme` include from this branch are all in `main` unchanged. What #767 added was the placement fix: the check had to move out of Quarto's `post-render` (which fires per-page when `build.sh --render-all` renders page-by-page, so every page saw the not-yet-rendered ones and failed) into a single Step 4b after the render loop.

Recording the credit: your refusal to "fix" a caption malformation that did not exist — and building an xml2 DOM-level audit to check my claim rather than accepting it — was correct. My diagnosis was a regex over raw HTML matching DT's embedded script JSON. The 618-word length was the real defect and your restructure fixed it.

Refs #767.

Issue #728, opened 2026-08-23
The leaderboard's statistical-rigour columns cover 2-9 of 17 strategies — and only 1 of the 8 claiming positive Sharpe has any multiple-testing correction


JohnGavin:
Found while implementing #726 item 4 (Bonferroni-correct the detection-power
requirement via `k_eff_strat`). The correction turned out to be nearly inert, and the
reason generalises well beyond that one column.

## The measurement

Full Period rows, current store (771 targets, 0 errored):

| column | non-NA rows |
|---|---|
| `hac_sharpe` | **2** / 17 |
| `wf_corr` | **2** / 17 |
| `deflated_sharpe` | **4** / 17 |
| `dsr_pvalue` | **4** / 17 |
| `k_eff_strat` | **4** / 17 |
| `sharpe_ci_lo` / `_hi` | **4** / 17 |
| `incremental_sharpe` | **5** / 17 |
| `ssr` | **9** / 17 |
| `detection_min_n_years` | 7 / 17 (+9 correctly NA on negative Sharpe) |

Every one of the multiple-testing columns — `deflated_sharpe`, `dsr_pvalue`,
`k_eff_strat` — is populated for exactly four strategies: Factor MAX, Factor DRIF,
Stock MAX, Stock DRIF. They come through `strat_deflated_sharpe`'s `col_map`, which
covers that family and nothing else.

## Why this matters more than it looks

**Of the eight strategies claiming a positive Full-Period Sharpe, exactly one
(Factor DRIF) has a deflated Sharpe or a `k_eff_strat`.** The other seven — LTR,
OLMAR-1, Risk State, Avoid Worst, Mom Pre-Peak, Value (HML), Managed Futures — carry
a raw Sharpe with no multiple-testing correction of any kind.

That is the wrong seven to be missing. These are the rows a reader is most likely to
act on.

It also makes #726 item 4 mostly cosmetic as specified: Bonferroni-correcting alpha
by `k_eff_strat` can only fire where `k_eff_strat` exists, so the corrected
detection-power columns would be NA for 7 of the 8 rows they were meant to
discipline. PR #727 implements it correctly and flags this; the fix belongs here, not
there.

## The deeper problem is the multiplicity count itself

`k_eff_strat = 4.847` is the effective number of tested strategies *within the
factor/stock MAX/DRIF family*. But we run **seventeen** strategies against overlapping
data and rank them against each other on one leaderboard. The multiplicity that
[...]

Issue #643, opened 2026-08-05
Value (HML) and Managed Futures are silently missing from the headline ranking — inconsistent period vocabulary


JohnGavin:
Found while verifying the re-rendered leaderboard after #642: the main ranking table shows **15** strategies, not the 17 that have full-sample rows.

## The defect

Period labels are not on a common vocabulary. From `tar_read(leaderboard)` on `main` at `758cba7`:

```
       Full  Full Period   OOS  Testing  Training  Validation
          2           15     2       11        13           7
```

The two `Full` rows are **`Value (HML)`** and **`Managed Futures`**. Every other strategy uses `Full Period`. The same pair uses `OOS` where everything else uses `Testing`.

The headline ranking table filters `period == "Full Period"` (`docs/leaderboard.qmd:58`), so both strategies are **silently dropped from the project's central output**. Same filter at `:651` (structural breaks) drops them again. The partition table uses `period != "Validation"` and so does include them, which is why `Managed Futures` appears in the rendered HTML but not in the ranking.

## Root cause — the same one as #637

The `.norm_*` helpers in `R/plan_leaderboard.R` **rename columns but do not normalise vocabularies**. #637 was that same failure applied to *units* (`cagr`/`vol`/`max_dd` mixing fractions and percentages because no helper rescaled). This is the identical failure applied to *categorical values*: `.norm_value()` and `.norm_mf()` pass their source targets' period labels straight through.

`R/plan_ev_ebit.R` and `R/plan_managed_futures.R` emit `Full`/`OOS`; the shared partition vocabulary defined in `R/plan_partitions.R` is `Training`/`Testing`/`Validation` plus `Full Period`. Nothing reconciles them.

## Why the existing gate missed it

`qa_leaderboard_coverage` asserts *"all 16 strategies present in leaderboard"* — presence **anywhere** in the table, not presence **per period**. A strategy can be absent from every period-filtered view and still pass. The gate is satisfied while the headline output is incomplete.

## Impact

- Two live strategies are invisible in the ranking readers treat as authoritative, with no indication anything was excluded.
- `Managed Futures` is the only strategy with a genuine vol-target and leverage cap, so its absence also distorts the leverage picture in #635 — the implied-gross table there is computed from strategies that *do* appear.
- Any future consumer filtering on `period == "Full Period"` inherits the same silent gap.

## Fix

[...]

PR #770, opened 2026-08-27
fix(dashboards): collapse the two leaderboard captions 0b412a9c missed

JohnGavin:
## Summary

- Diagnosed the "leaderboard caption is still long" report using the widget-JSON extraction method (not DOM, not source grep) against both a standalone repro and the real, committed `docs/leaderboard.html`.
- **Finding: `hd_caption()`'s collapse for the Rankings and Financing Sensitivity captions is already working correctly** (visible portion 50/51 words; full disclosures behind a native `<details>` element, closed by default per HTML5 spec — confirmed against DT 0.34.0's JS source). The prior commit's claim that the "restructure failed silently" measured the *total* widget-JSON field length rather than the visible-by-default portion — the same DOM-vs-JSON confusion this dispatch was warned about, one level deeper.
- **What was actually broken:** two of leaderboard.qmd's five DT captions ("By Partition", 511 chars; "Structural Breaks", 1170 chars) were never migrated to `hd_caption()` — still built with raw `tags$caption()`, fully visible, no collapse. Both fixed now.
- Replaced a flawed length-audit test (measured total caption length, so it silently skipped on the real fixed HTML) with two tests that measure the *visible* portion specifically, plus a new render-independent source-level test that walks the actual R parse tree of `leaderboard.qmd` and would have caught both defects with no store and no render.
- Audited all 15 dashboard HTML files for the same defect class — found similar un-collapsed long captions in `bdbb-sol.html`, `european-overlay.html`, `falsification.html` (11 instances), `momentum-prepeak.html`. Out of scope for this dispatch's write-paths; flagging for a follow-up issue.
- Did not reach the bonus item (strategy-name links / hover detail on the leaderboard table) — the diagnosis above took the full time budget.

## Test plan

- [x] `docs/leaderboard.qmd` R chunks parse cleanly (`knitr::purl()` + `parse()`)
- [x] `tests/testthat/test-hd-caption.R` — 25/25 pass
- [x] `tests/testthat/test-dashboard-caption-wellformedness.R` — new/modified tests pass (2 expected skips against the not-yet-regenerated `docs/leaderboard.html`, documented pending-regen by the main checkout)
- [ ] `scripts/verify.sh` full suite — running in foreground, will report in a follow-up comment
- [ ] Once merged, the main checkout should regenerate `docs/leaderboard.html`/`docs/_targets` and re-run this PR's two now-skipped HTML-parsing tests to confirm they PASS against fresh output

Refs #767, #728, #726, #643, #637

### DIFF
--- a/docs/leaderboard.qmd
+++ b/docs/leaderboard.qmd
@@ -645,13 +645,24 @@ if (!is.null(lb)) {
   numeric_cols <- which(sapply(by_part, function(x) is.numeric(x) ||
                                   grepl("^[0-9.%$,-]+$", as.character(x[1]))))
 
-  DT::datatable(
-    by_part,
-    caption = htmltools::tags$caption(
-      style = "caption-side: top; text-align: left; font-weight: bold;",
-      class = "dt-caption",
+  # 511 chars (> hd_caption()'s 300-char threshold) but was a raw
+  # tags$caption() with no collapse -- audited and fixed alongside the
+  # Rankings/financing-sensitivity captions (dispatch 2d451d09, follow-up to
+  # 0b412a9c which fixed those two but missed this one and Structural Breaks
+  # below).
+  by_part_caption <- hd_caption(
+    paste0(
       "All strategies × Training, Testing, and Holdout partitions, ordered Testing / Holdout / Full Period / Training then by Sharpe (descending). Years column shows date ranges. Holdout (2024-2026) is observed, not sealed -- treat it with a discount relative to a genuinely untouched sample (#660). Validation partition is sealed and genuinely untouched (2026-05-01 onwards) — not shown until production deployment."
     ),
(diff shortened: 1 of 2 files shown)


JohnGavin:
`scripts/verify.sh` (full mode, foreground): **PASS**

```
SKIP count this run: 2

SKIP count this run: 15

```

The 2 root-suite skips are this PR's new pending-regen skips in `test-dashboard-caption-wellformedness.R` (documented in-test, referencing this dispatch) — they will flip to PASS once the main checkout regenerates `docs/leaderboard.html`. No new failures; both baselines matched exactly.

`scripts/check_dashboard_freshness.R` (Check 2) correctly flags `leaderboard.qmd` as source-newer-than-.html (advisory, does not fail the run) — expected, since this PR only changes the `.qmd`/test source, not the rendered artifact.

`docs/leaderboard.qmd` R chunks confirmed to parse cleanly via `knitr::purl()` + `parse()`.

PR #773, opened 2026-08-27
fix(roborev): route reviews to claude-code — correctness and privacy

JohnGavin:
Routes this repo's roborev reviews to `claude-code` instead of inheriting the global `gemini` default. Two independent reasons, either sufficient alone.

## 1. Correctness — gemini cannot read the diff it is asked to review

roborev writes its snapshot diff *inside the repo under review* at `<repo>/.roborev/roborev-snapshot-*/roborev-snapshot-content.diff`. `.roborev` is gitignored here, so gemini answers:

> I am unable to access the diff file `…` because it is ignored by configured ignore patterns.

…and produces no review — while the job records `status=done` and `verdict=0`. **The commit is effectively unreviewed but looks reviewed-and-clean.**

Measured 2026-08-27 against `~/.roborev/reviews.db` (open reviews), this repo:

| agent | reviews | could-not-read |
|---|---|---|
| gemini | 25 | **6** |
| claude-code | 6 | **0** |

Same repo, same gitignored path. The gitignore is necessary but **not sufficient** — the refusal is agent behaviour. Across all repos: gemini 20/129 (15.5%), claude-code 0/36.

## 2. Privacy — the stronger reason here

This repo holds personal financial data. Per the `public-private-repo-boundary` rule, **money** is one of the four triggers that makes content private.

`llm/.roborev.toml` states the intended policy explicitly:

> GLOBAL default_agent stays 'claude-code' so private repos (no per-repo config) never route to gemini — public-only opt-in.

…and justifies gemini for `llm` alone on the grounds that *"llm is PUBLIC so free-tier 'trains on prompts' is moot"*. That reasoning does not transfer to a private finance repo.

**The global default has since drifted:** `~/.roborev/config.toml` now has `default_agent = 'gemini'`. This repo was inheriting exactly the routing the policy was written to prevent. Setting it per-repo restores the documented intent regardless of what the global default does next.

> Worth noting separately: `premortem` (estate) and `mycare` (NHS health) are also currently routed to gemini by the same drift. Those are not fixed by this PR.

## The model pins are not optional

Global config pins `review_model` and `review_model_thorough` to `gemini-2.5-flash-lite`. Those are **model** pins, not agent pins, so setting the agent alone would leave claude-code inheriting a gemini model name and failing — the `llm#746` "poison". Every review-model key this repo can reach is pinned to `sonnet`.

## No gemini backup here — deliberately

Unlike the public `llmtelemetry`, this repo sets `review_backup_agent = 'claude-code'`. Falling back to gemini would send private financial diffs to the agent this config exists to keep them away from. If claude-code is unavailable, the correct outcome is a **failed review someone notices**, not a completed review by the wrong agent.

## Verification

Enqueued a real review of HEAD (`job 12797`), polled to `status=done`, and confirmed the stored output classifies as **"reviewed, findings parsed"** (2088 chars) rather than a could-not-read refusal. `roborev review` reported `agent: claude-code` at enqueue.

Output content was deliberately not inspected or quoted — private repo; only the classification was read.

Refs [JohnGavin/llm#1035](https://github.com/JohnGavin/llm/issues/1035)

### DIFF
--- a/.roborev.toml
+++ b/.roborev.toml
@@ -33,3 +33,50 @@ exclude_patterns = [
 #    `docs/` as a code-review concern — it is the deployed artifact.
 #    Internal-only HTML (e.g. examples.html, dashboard.html dev builds) IS
 #    gitignored via docs/.gitignore where appropriate.
+
+# ── Review agent: claude-code, not the global gemini default (llm#1035) ──────
+#
+# WHY (correctness): gemini refuses to read roborev's own snapshot diff. roborev
+# writes it to <repo>/.roborev/roborev-snapshot-*/roborev-snapshot-content.diff
+# and `.roborev` is gitignored here, so gemini answers "I am unable to access
+# the diff file ... because it is ignored by configured ignore patterns" and
+# produces no review — while the job records status=done and verdict=0, so the
+# commit looks reviewed-and-clean when it was never reviewed at all.
+#
+# Measured 2026-08-27 on ~/.roborev/reviews.db (open reviews), THIS repo:
+#   gemini        25 reviews,  6 could-not-read
+#   claude-code    6 reviews,  0 could-not-read
+# Same repo, same gitignored path — so the gitignore is necessary but not
+# sufficient; the refusal is agent behaviour. Across all repos it is 20/129
+# (15.5%) for gemini and 0/36 for claude-code.
+#
+# WHY (privacy — the stronger reason for THIS repo): this repo holds personal
+# financial data. Per the public-private-repo-boundary rule, "money" is one of
+# the four triggers that makes content private. llm/.roborev.toml states the
+# intended policy explicitly: "GLOBAL default_agent stays 'claude-code' so
+# private repos (no per-repo config) never route to gemini — public-only
+# opt-in", and justifies gemini for llm only because "llm is PUBLIC so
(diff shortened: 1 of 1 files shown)


JohnGavin:
**Correction to this PR's rationale** (the change itself stands; one justification in it was overstated).

The PR body and commit message say the model pins "are not optional" / "MUST accompany the agent change", on the grounds that the global `review_model = 'gemini-2.5-flash-lite'` would poison a claude-code agent override.

Measured afterwards, that is not what happens. `crypto_solwatch` pins `agent = "claude-code"` and **no** model, and its jobs are fine:

```
status  model    n   with_error
done    (none)   15  0
```

So a repo-level agent override alone is sufficient; the global model pin does not leak into it. The documented `llm#746` poison is narrower than I described — it is a *primary's* model pin leaking into the **backup** agent when the backup model keys are empty, which is a different path.

The model pins in this PR are therefore **belt-and-braces, not load-bearing**: they make the resolved model explicit rather than inherited, which is still worth having, but the agent line is what actually fixes the problem. No change to the diff — only to the reasoning attached to it.

Flagging rather than quietly leaving the stronger claim in the commit history, since a future reader would otherwise treat the model pin as mandatory and copy it as such.

